PR #7, opened 2018-03-09
Fix incorrect IncompatibleClassChangeError in ClassInfo.getDefaultMethod

upthewaterspout:
The logic to detect if there are two conflicting default methods was incorrect.
If a class and its superclass both implement the same interface JPF should not
throw an IncompatibleClassChangeError.


cyrille-artho:
Thanks for your contribution. Sorry for the slow action; we were busy setting up GSoC, but now we have Travis ready for future pull requests, so we can accept them much faster in the future.

PR #12, opened 2018-04-17
Removing references to JavaObjectInputStreamAccess 

wvisser:
It is no longer in Java8_u161


wvisser:
The problem is that once travis updates to java8_u161 then the master BUILD will fail. So it is a trade-off between the build failing for everybody on u161 and the test failing for the stragglers still on u151 (and everybody before u144). Can one tell travis which version of 1.8 to use?

We should merge the travis config also into dev.

gayanW:
@wvisser Yeah I understand. Thanks for the information. 

Let's upgrade it to use the latest available oracle-jdk that is provided in the PPA [oracle-java8-installer](https://launchpad.net/%7Ewebupd8team/+archive/ubuntu/java).

I'll test and create a PR for this.

gayanW:
@wvisser PR. #13

wvisser:
Okay shockingly JavaObjectInputStreamAccess is back in u171 that was released on April 17th. I'll scrap this pull request, since it is no longer required.

PR #13, opened 2018-04-17
Make Travis build against latest available oracle-jdk 

gayanW:
jpf-core requires a newer release of Oracle JDK than the pre-installed version provided by the Travis. This will make Travic CI to use the latest available oracle-jdk that is provided from the PPA oracle-java8-installer.

References:
https://docs.travis-ci.com/user/languages/java/#Updating-Oracle-JDK-to-a-recent-release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,17 @@
 language: java
 before_install:
-  - sudo apt-get install junit4
   - mkdir ~/junit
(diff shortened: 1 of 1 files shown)


wvisser:
Annoyingly it seems Oracle has an issue with the downloading. The build thus fails, but I don't think we broke anything, we just need to wait for Oracle to fix the installer download

Issue #17, opened 2018-04-27
.travis.yml to build against Java 10 


gayanW:
Modify .travis.yml so that it builds against Oracle Java 10 instead of Oracle Java 8. This is to support Java 10/11 and may go in a separate branch (eg. java-10)

References:
https://docs.travis-ci.com/user/getting-started/
https://github.com/javapathfinder/jpf-core/wiki/GSoC-2018-Project-Ideas#support-java-9-for-jpf-core

PR #18, opened 2018-04-27
Modify .travis.yml to build against java-10 (#17)

gayanW:
Modify .travis.yml so that it builds against the latest available oracle-jdk 10 that is provided from the PPA linuxuprising/java.

Upgrade Ant as module compilation is only available since Ant 1.9.7

We should ideally use the webupd8team/java PPA as suggested by travis docs. But jdk-10 is not yet available there, hence using ppa:linuxuprising/java

Fixes: #17

References:
https://docs.travis-ci.com/user/languages/java/#Updating-Oracle-JDK-to-a-recent-release
https://docs.travis-ci.com/user/installing-dependencies/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,26 @@
 language: java
+
+addons:
+  apt:
+    sources:
+      - sourceline: 'ppa:linuxuprising/java'
+    packages:
+      - oracle-java10-installer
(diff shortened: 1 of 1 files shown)


gayanW:
Closing because the PR should be created against a new branch instead of `master`.

PR #45, opened 2018-05-16
Added minimal Gradle support to jpf-core

### LINKED ISSUE #43: Add minimal Gradle support to jpf-core
To kick off the migration to Gradle, the project must have a minimal Gradle support.
Concretely, this means being able to use Gradle as an interface to Ant tasks.
To this end, I need to:

1. Add the Gradle wrapper (currently targeting version 4.7, as it has some interesting features, including Java 10 support - https://docs.gradle.org/4.7/release-notes.html)
2. Import the build.xml script, so Ant tasks can be called via Gradle

jeandersonbc:
Minimal Gradle support consists in:

1) Gradle wrapper configured;
2) Minimal working build script

Note that testing is not currently supported but it will be supported as I evolve the Gradle support.
This Pull Request closes #43


jeandersonbc:
Curiously, TravisCI attempted to run the build with Gradle automatically. Seems like we will have a broken CI in the gradle branch for a while

cyrille-artho:
Could you please add a target "check" that succeeds, so Travis will not report a failure?
It's OK if we don't run the unit tests for the time being, but I would like to have a passing CI check if possible, so we know that the build works as expected.

jeandersonbc:
All good now. The Ant build is running before the dummy Gradle build.

PR #49, opened 2018-05-20
Disable Xlint:deprecation when compiling sources

gayanW:
Temporarily disable Xlint:deprecation just to reduce the verbosity of the build logs.

eg:-  warning: [deprecation] Integer(int) in Integer has been deprecated

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -25,7 +25,7 @@
   
   <!-- compiler settings -->
   <property name="debug"         value="on"/>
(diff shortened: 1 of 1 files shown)


cyrille-artho:
I would prefer to leave the warnings, but feel free to disable them locally (on your machine).

gayanW:
Sure. I'll try fix the warnings instead.

Issue #50, opened 2018-05-21
error: [junit] cannot load system class java.lang.Object


gayanW:
When running tests compiled with Java 9 and later, results in the run time errors look like the following:

    [junit] Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
    [junit] Running TypeNameTest
    [junit]   running jpf with args:
    [junit] [SEVERE] cannot load system class java.lang.Object
    [junit] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.521 sec
    [junit] gov.nasa.jpf.vm.ClassInfoException: class not found: java.lang.Object
    [junit] 	at gov.nasa.jpf.vm.ClassLoaderInfo.getResolvedClassInfo(ClassLoaderInfo.java:363)
    [junit] 	at gov.nasa.jpf.vm.SystemClassLoaderInfo.getResolvedClassInfo(SystemClassLoaderInfo.java:147)
    [junit] 	at gov.nasa.jpf.vm.VM.getStartupSystemClassInfos(VM.java:445)
    [junit] 	at gov.nasa.jpf.vm.VM.initializeMainThread(VM.java:564)
    [junit] 	at gov.nasa.jpf.vm.SingleProcessVM.initialize(SingleProcessVM.java:130)
    [junit] 	at gov.nasa.jpf.JPF.run(JPF.java:611)
    [junit] 	at gov.nasa.jpf.util.test.TestJPF.createAndRunJPF(TestJPF.java:675)
    [junit] 	at gov.nasa.jpf.util.test.TestJPF.noPropertyViolation(TestJPF.java:806)
    [junit] 	at gov.nasa.jpf.util.test.TestJPF.verifyNoPropertyViolation(TestJPF.java:830)
    [junit] 	at TypeNameTest.testArrayCloning(TypeNameTest.java:58)

Possibly because of the new Module System, the class URI of the generated  .class files in the build destination directory are different than previously, which makes JPF classloader to fail resolve the model classes from jpf-classes.jar

JDK 8

    build/classes/
    ├── java
    │   ├── io
    │   ├── lang
    │   │   ├── annotation
    │   │   ├── ref
    │   │   └── reflect
    
JDK 9/10
    
    build/classes/
    ├── java.base
    │   ├── java
    │   │   ├── io
    │   │   ├── lang
    │   │   │   ├── annotation
    │   │   │   ├── ref
    │   │   │   └── reflect

[...]

PR #51, opened 2018-05-21
Rearrange class files in jpf-classes.jar

gayanW:
Before Java module system (introduced in JDK 9), once compiled, all compiled classes are nested inside the destination directory, based on their respective packages. So for instance,

```
build/classes/
├── java
│   ├── io
│   ├── lang
│   │   ├── annotation
│   │   ├── ref
│   │   └── reflect
```

But when compiled with Java 9/10, directory structure of the destination directory for the generated class files is as follows.

```
build/classes/
├── java.base
│   ├── java
│   │   ├── io
│   │   ├── lang
│   │   │   ├── annotation
│   │   │   ├── ref
│   │   │   └── reflect
```

So if we are to include classes from build/classes directory as a fileset into jpf-classes.jar, as done previously, like:
    <jar jarfile="build/jpf-classes.jar">
      <fileset dir="build/classes"/>

.. JPF classloader will fail to resolve the model classes, because the class urls are different than previously.

```
[junit] gov.nasa.jpf.vm.ClassInfoException: class not found: java.lang.Object
 at gov.nasa.jpf.vm.ClassLoaderInfo.getResolvedClassInfo(ClassLoaderInfo.java:363)
 at gov.nasa.jpf.vm.SystemClassLoaderInfo.getResolvedClassInfo(SystemClassLoaderInfo.java:147)
 [SEVERE] cannot load system class java.lang.Object
```

So this restructure the files in jpf-classes.jar (same as before) without introducing any implementation changes to the code.

Fixes: #50

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -197,10 +197,6 @@
     <copy file=".version" todir="build/main/gov/nasa/jpf" failonerror="false"/>
 
     <jar jarfile="build/jpf-classes.jar">
-      <fileset dir="build/classes">
-        <include name="gov/nasa/jpf/*.class"/>
-        <include name="org/junit/*.class"/>
-      </fileset>
       <fileset dir="build/classes/java.base"/>
       <fileset dir="build/classes/java.logging"/>
 


gayanW:
I added a new commit that excludes  "gov/nasa/jpf/*.class", and "org/junit/*.class" class files from jpf-classes.jar

PR #60, opened 2018-05-28
Revert "Format output of ant 'test' target"

gayanW:
Reverts javapathfinder/jpf-core#48

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -444,8 +444,8 @@
       </not>
     </condition>
     
-    <junit printsummary="on" showoutput="on"
-           haltonfailure="yes" logfailedtests="true" failureproperty="test.failed"
+    <junit printsummary="on" showoutput="off" 
+           haltonfailure="no" logfailedtests="true" failureproperty="test.failed" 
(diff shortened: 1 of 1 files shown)


gayanW:
@cyrille-artho Sure I'll look into it. 

In our JUnit task printsummary is set to on. However as the `forkmode` is set to `perTest`, test result summary is printed per test. I'll see if there's a workaround to get it to output the final result.

https://github.com/javapathfinder/jpf-core/blob/a83107dbb0e5959d13c4be1121afed37cfa8c74a/build.xml#L447-L449

cyrille-artho:
In that case, why not just turn off forkmode.
We can afford to lose some time, and we'll try to get the best and most efficient configuration later, for the gradle-supported build.

gayanW:
It looks like forking is necessary. When turned off, failure of one test seems to cause all subsequent tests fail.

As for the test outputs what about generating a HTML report for Junit results? In that case we could inspect log outputs, errs per each test in a browser locally without cluttering the console output. I think it would be good to do it in a separate PR.

cyrille-artho:
OK, then let's leave the configuration as by this PR.
There is no need to look into this further, as the gradle build will already updated the test output.

Issue #81, opened 2018-06-07
warning: Reflection is internal proprietary API and may be removed


gayanW:
Deprecation warning 

    'Reflection is internal proprietary API and may be removed in a future release' 

when build with JDK 9 and later:

    [javac] /home/travis/build/javapathfinder/jpf-core/src/tests/gov/nasa/jpf/test/vm/reflection/ReflectionTest.java:34: warning: Reflection is internal proprietary API and may be removed in a future release
    [javac]       Class<?> callerCls = sun.reflect.Reflection.getCallerClass(0); // that would be getCallerClass()
    [javac]                                       ^

Travis log:
https://travis-ci.org/javapathfinder/jpf-core/builds/389306467#L2485-L2508

References:
https://www.logicbig.com/tutorials/core-java-tutorial/java-9-changes/stack-walker-caller.html

gayanW:
If we are to use the StackWalker instead of calling sun.reflect.Reflection.getCallerClass, I think it will make sense to refactor the name of the test class from 'gov.nasa.jpf.test.**vm.reflection.ReflectionTest**' to 'gov.nasa.jpf.test.**java.lang.StackWalkerTest**'

For your reference:
https://github.com/javapathfinder/jpf-core/blob/18a0c42de3e80be0c2ddcf0d212e376e576fcda0/src/tests/gov/nasa/jpf/test/vm/reflection/ReflectionTest.java#L24-L36

cyrille-artho:
Yes, please update the class name for consistency.

PR #83, opened 2018-06-13
Move/Refactor ReflectionTest -> StackWalkerTest

gayanW:
`sun.reflect.Reflection#getCallerClass` is an internal API and is marked
for removal. This uses StackWalker to walk the stack and obtain the
caller class with `StackWalker.StackFrame.getDeclaringClass` instead.

The class is also being renamed from
'gov.nasa.jpf.test.vm.reflection.ReflectionTest' to
'gov.nasa.jpf.test.java.lang.StackWalkerTest' for consistency.

Fixes: #81


gayanW:
I remote debug to make sure that the assertions in this test pass, as the test would still fail when running `ant test` because of other unresolved issues which are not relevant to this issue.

PR #84, opened 2018-06-14
Added missing packaging tasks on Gradle build

jeandersonbc:


cyrille-artho:
The patch seems functionally correct, but there is a lot of duplication.
Can you please check if you can set version and other variables in an intermediate step that can be shared, and if it is possible to share include settings?

cyrille-artho:
OK, we can do this as a clean-up. In that case, I'll merge this PR and close its corresponding issue; please make a new issue for the clean-up.

Issue #90, opened 2018-06-16
warning: auxiliary class CompoundEnumeration in ClassLoader.java should not be accessed from outside its own source file


gayanW:
Following compiler warning shows up in the build logs:

    [javac] src/classes/modules/java.base/java/lang/ClassLoader.java:112: warning: auxiliary class CompoundEnumeration in ClassLoader.java should not be accessed from outside its own source file
    [javac]     return new CompoundEnumeration<URL>(resEnum);
    [javac]                ^

Travis log:
https://travis-ci.org/javapathfinder/jpf-core/builds/393023635#L2444-L2446

Octarine-J:
This warning is shown when accessing a non-public class from a .java file that contains multiple classes. 

Although it's not recommended, it is possible to declare multiple classes in one .java file. However, only one of those classes can be public:

public class A {
   // ...
}

class B {
  // this is an auxiliary class
}

gayanW:
The current code expects to have java.lang.CompoundEnumeration public standard Java class. But due to changes in the JDK, there seems to have no such public class. In JDK 9/10 for example it is declared in the java/lang/ClassLoader.java itself.

Please see the PR.

https://github.com/javapathfinder/jpf-core/blob/cf154ec4212dbdffa8cd7345a7abfa13d4dd18e4/src/classes/modules/java.base/java/lang/ClassLoader.java#L102-L113

References:
http://hg.openjdk.java.net/jdk10/jdk10/jdk/rev/47e7d7363249

Octarine-J:
It seems that this class is a merely utility that allows continuous iteration over all values from an array of enumerations. The class name is probably not important; the essential part lies in the behaviour. Maybe we can rewrite this method to return an instance of Enumeration that contains all values from the arrays resEnum[0] and resEnum[1]?

By the way, this Enumeration class is from JDK 1.0, and probably is not used widely anymore; its name is a bit unfortunate as one can confuse it with Enum. This Enumeration<E> class seems to be a less poverful equivalent of Iterator<E>.

gayanW:
Thanks for the tip. What about this implementation, though it is quite verbose?

```
  public Enumeration<URL> getResources(String name) throws IOException {
    Vector<URL> urlVector = new Vector<>(2);
    if(parent == null) {
      Iterator<URL> iterator = getSystemClassLoader().getResourcesURL(name).asIterator();
      while (iterator.hasNext()) {
        urlVector.add(iterator.next());
      }
    } else{
[...]

PR #91, opened 2018-06-18
Add new auxliary class CompoundEnumeration to ClassLoader.java

gayanW:
Our java.lang.ClassLoader model class references
java.lang.CompoundEnumeration which is an auxilary class declared within
the java standard class java/lang/ClassLoader.java

This includes CompoundEnumeration class within our model class itself,
so to prevent our ClassLoader model class from referencing an auxilary
class from outside its own source file.

This fixes:
    warning: auxiliary class CompoundEnumeration in ClassLoader.java should
             not be accessed from outside its own source file

Fixes: #90

### DIFF
--- a/src/classes/modules/java.base/java/lang/ClassLoader.java
+++ b/src/classes/modules/java.base/java/lang/ClassLoader.java
@@ -24,6 +24,7 @@ import java.net.URL;
 import java.nio.ByteBuffer;
 import java.security.ProtectionDomain;
 import java.util.Enumeration;
+import java.util.NoSuchElementException;
 import java.util.Vector;
 
 /**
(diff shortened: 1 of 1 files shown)


cyrille-artho:
So to clarify, this code used to be compiled against the library of the host JVM?
In that case, if we use the code from openjdk, we'll have to make sure the licenses are compatible. Can we have this enumeration class in a separate file?

gayanW:
> So to clarify, this code used to be compiled against the library of the host JVM?

We don't maintain CompoundEnumeration in jpf-core. ClassLoader model class resolves CompoundEnumeration from the same class file ClassLoader.java but from the library of the host JVM. 

Yes it is possible to put it in a separate class file. In that case is to okay to having a model class without corresponding Java replacement. BTW why we'd need it to put in a separate file?

gayanW:
Our ClassLoader model class resolves CompoundEnumeration from the same class file ClassLoader.java but from the library of the host JVM. 

Yes it is possible to put it in a separate class file. In that case is to okay to have a model class for which the JDK counterpart does not exist (ie. there is no CompoundEnumeration.java in OpenJDK as its declared as an auxiliary class in ClassLoader.java). Why we'd need it to put in a separate file? Is it because of the licensing issue?

Octarine-J:
Actually, I have never considered the fact that licenses of OpenJDK and JPF may be incompatible. I am not very familiar with open source licenses; maybe we can discuss this on the meeting? I guess, in the worst case we can just ignore this warning (unless there are many others of the same kind).

gayanW:
It is not possible to suppress it with SuppressWarnings annotation either. I tried annotating the method with `SuppressWarnings("all")`, the warning appears still.

cyrille-artho:
Ok, let's just add a comment and close this issue with that.

--
Regards,
Cyrille Artho

Issue #109, opened 2018-07-21
NoSuchMethodException:SharedSecrets.setJavaNetURLAccess(Ljdk/internal/misc/JavaNetURLAccess;)V


gayanW:
Runtime error _"java.lang.NoSuchMethodException: jdk.internal.misc.SharedSecrets.setJavaNetURLAccess(Ljdk/internal/misc/JavaNetURLAccess;)V"_ when running ClassLoaderTest.

    [junit] java.lang.NoSuchMethodException: jdk.internal.misc.SharedSecrets.setJavaNetURLAccess(Ljdk/internal/misc/JavaNetURLAccess;)V
    [junit]     at java.net.URL.<clinit>(URL.java:1632)
    [junit]     at java.lang.ClassLoader.getResource(ClassLoader.java:62)
    [junit]     at java.lang.ClassLoader.getResource(ClassLoader.java:67)
    [junit]     at java.lang.ClassLoader.getResourceAsStream(ClassLoader.java:131)
    [junit]     at gov.nasa.jpf.test.java.lang.ClassLoaderTest.testGetResourceAsStreamImpl(ClassLoaderTest.java:164)
    [junit]     at gov.nasa.jpf.test.java.lang.ClassLoaderTest.testGetResourceAsStream(ClassLoaderTest.java:51)
    [junit]     at java.lang.reflect.Method.invoke(gov.nasa.jpf.vm.JPF_java_lang_reflect_Method)
    [junit]     at gov.nasa.jpf.util.test.TestJPF.runTestMethod(TestJPF.java:648)

PR #110, opened 2018-07-21
Add accessor methods for SharedSecrets#javaNetURLAccess

gayanW:
Adds SharedSecrets#setJavaNetURLAccess and SharedSecrets#getJavaNetURLAccess methods to jdk.internal.misc.SharedSecrets model clsas.

This fixes:

    [junit] java.lang.NoSuchMethodException:
            jdk.internal.misc.SharedSecrets.setJavaNetURLAccess(Ljdk/internal/misc/JavaNetURLAccess;)V

Fixes: #109

### DIFF
--- a/src/classes/modules/java.base/jdk/internal/misc/SharedSecrets.java
+++ b/src/classes/modules/java.base/jdk/internal/misc/SharedSecrets.java
@@ -52,6 +52,7 @@ public class SharedSecrets {
   private static JavaUtilJarAccess javaUtilJarAccess;
   private static JavaLangAccess javaLangAccess;
   private static JavaIOAccess javaIOAccess;
+  private static JavaNetURLAccess javaNetURLAccess;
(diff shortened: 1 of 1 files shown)


src/classes/modules/java.base/jdk/internal/misc/SharedSecrets.java
```diff
@@ -77,6 +78,15 @@ public static JavaNetAccess getJavaNetAccess() {
...
   }
 
+  public static void setJavaNetURLAccess(JavaNetURLAccess jnua) {
+    javaNetURLAccess = jnua;
+  }
+
+  public static JavaNetURLAccess getJavaNetURLAccess() {
+    if (javaNetURLAccess == null)
+      unsafe.ensureClassInitialized(java.net.URL.class);
+    return javaNetURLAccess;
```
cyrille-artho: Is this method allowed to return null, or should it also initialize an instance?

Issue #111, opened 2018-07-21
NoSuchMethodException: SharedSecrets.setJavaObjectInputFilterAccess(Ljdk/internal/misc/JavaObjectInputFilterAccess;)V


gayanW:
Runtime error _"java.lang.NoSuchMethodException: jdk.internal.misc.SharedSecrets.setJavaObjectInputFilterAccess(Ljdk/internal/misc/JavaObjectInputFilterAccess;)V"_ when running ObjectStreamTest.

    [junit] java.lang.NoSuchMethodException: jdk.internal.misc.SharedSecrets.setJavaObjectInputFilterAccess(Ljdk/internal/misc/JavaObjectInputFilterAccess;)V
    [junit]     at java.io.ObjectInputFilter$Config.<clinit>(ObjectInputFilter.java:271)
    [junit]     at java.io.ObjectInputStream.<init>(ObjectInputStream.java:340)
    [junit]     at gov.nasa.jpf.test.java.io.ObjectStreamTest.testSimpleReadbackOk(ObjectStreamTest.java:88)
    [junit]     at java.lang.reflect.Method.invoke(gov.nasa.jpf.vm.JPF_java_lang_reflect_Method)
    [junit]     at gov.nasa.jpf.util.test.TestJPF.runTestMethod(TestJPF.java:648)

PR #112, opened 2018-07-21
Add accessor methods for SharedSecrets#javaObjectInputFilterAccess

gayanW:
Adds SharedSecrets#getJavaObjectInputFilterAccess and SharedSecrets#setJavaObjectInputFilterAccess methods to jdk.internal.misc.SharedSecrets model class.

This fixes:

```
[junit] java.lang.NoSuchMethodException:
        jdk.internal.misc.SharedSecrets.setJavaObjectInputFilterAccess(Ljdk/internal/misc/JavaObjectInputFilterAccess;)V
```
Fixes: #111

### DIFF
--- a/src/classes/modules/java.base/jdk/internal/misc/SharedSecrets.java
+++ b/src/classes/modules/java.base/jdk/internal/misc/SharedSecrets.java
@@ -19,6 +19,7 @@ package jdk.internal.misc;
 
 import java.io.File;
 import java.io.FileDescriptor;
+import java.io.ObjectInputFilter;
 import java.io.ObjectInputStream;
 import java.util.jar.JarFile;
 
@@ -58,6 +59,7 @@ public class SharedSecrets {
   private static JavaNioAccess javaNioAccess;
   private static JavaAWTAccess javaAWTAccess;
   private static JavaObjectInputStreamAccess javaObjectInputStreamAccess;
+  private static JavaObjectInputFilterAccess javaObjectInputFilterAccess;
 
   // (required for EnumSet ops)
   public static JavaLangAccess getJavaLangAccess() {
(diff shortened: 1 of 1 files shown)


src/classes/modules/java.base/jdk/internal/misc/SharedSecrets.java
```diff
@@ -150,7 +152,18 @@ public static JavaObjectInputStreamAccess getJavaObjectInputStreamAccess() {
   public static void setJavaObjectInputStreamAccess(JavaObjectInputStreamAccess access) {
     javaObjectInputStreamAccess = access;
   }
-  
+
+  public static JavaObjectInputFilterAccess getJavaObjectInputFilterAccess() {
+    if (javaObjectInputFilterAccess == null) {
+      unsafe.ensureClassInitialized(ObjectInputFilter.Config.class);
+    }
+    return javaObjectInputFilterAccess;
```
cyrille-artho: Same as with the previous issue: should this code initialize a singleton, or possibly return null, if (javaObjectInputFilterAccess == null)?

gayanW: Our ensureClassInitialized method seems to do nothing. So it will possibly return null.

https://github.com/javapathfinder/jpf-core/blob/09c442cdcdc364aa65fcae08a506b749c47969c5/src/classes/modules/java.base/jdk/internal/misc/Unsafe.java#L215

https://github.com/javapathfinder/jpf-core/blob/09c442cdcdc364aa65fcae08a506b749c47969c5/src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java#L260-L264

JDK source for ensureClassInitialized:
https://github.com/unofficial-openjdk/openjdk/blob/6caf29475edb2d25cb9349f763451e20aa545c74/src/java.base/share/classes/jdk/internal/misc/Unsafe.java#L1032-L1043

cyrille-artho: I see, this is indeed the same as in the JDK implementation. Merging this one.

PR #125, opened 2018-07-27
Add java.lang.Class#getModule to MJI model class

### LINKED ISSUE #124: NoSuchMethodError: java.lang.Class.getModule()Ljava/lang/Module
Runtime error _"java.lang.NoSuchMethodError: java.lang.Class.getModule()Ljava/lang/Module"_ when running CountDownLatchTest.

    [junit] java.lang.NoSuchMethodError: java.lang.Class.getModule()Ljava/lang/Module;
    [junit]     at sun.invoke.util.VerifyAccess.isClassAccessible(VerifyAccess.java:180)
    [junit]     at java.lang.invoke.MethodHandles$Lookup.checkSymbolicClass(MethodHandles.java:2047)
    [junit]     at java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:2020)
    [junit]     at java.lang.invoke.MethodHandles$Lookup.findVarHandle(MethodHandles.java:1516)
    [junit]     at java.util.concurrent.locks.AbstractQueuedSynchronizer.<clinit>(AbstractQueuedSynchronizer.java:2301)
    [junit]     at java.util.concurrent.CountDownLatch.<init>(CountDownLatch.java:201)
    [junit]     at gov.nasa.jpf.test.java.concurrent.CountDownLatchTest.testCountDown(CountDownLatchTest.java:43)
    [junit]     at java.lang.reflect.Method.invoke(gov.nasa.jpf.vm.JPF_java_lang_reflect_Method)
    [junit]     at gov.nasa.jpf.util.test.TestJPF.runTestMethod(TestJPF.java:648)

gayanW:
Add method java.lang.Class#getModule to  MJI model class for java.lang.Class

This also fixes #124 :

```
[junit] java.lang.NoSuchMethodError:
        java.lang.Class.getModule()Ljava/lang/Module;
```

### DIFF
--- a/src/classes/modules/java.base/java/lang/Class.java
+++ b/src/classes/modules/java.base/java/lang/Class.java
@@ -60,6 +60,9 @@ public final class Class<T> implements Serializable, GenericDeclaration, Type, A
   
   private String name;
 
+  // set by VM
+  private transient Module module;
+
   private ClassLoader classLoader;
   
   /**
@@ -358,5 +361,9 @@ public final class Class<T> implements Serializable, GenericDeclaration, Type, A
   public boolean isSynthetic (){
     final int SYNTHETIC = 0x00001000;
     return (getModifiers() & SYNTHETIC) != 0;
+  }
+
+  public Module getModule() {
+    return module;
   }
 }


src/classes/modules/java.base/java/lang/Class.java
```diff
@@ -60,6 +60,9 @@
   
   private String name;
 
+  // set by VM
```
cyrille-artho: Could you please clarify this comment?
Does JPF's VM or the host VM try to set this field?
There is currently no accessor method, and as the field is now, it seems jpf-core would not use it at this point. However, without an accessor, the host VM could not set this field (except by reflection, which can access private data).

gayanW: Based on the JDK source there is no accessor `setModule`. But through remote debugging, I was able to confirm that the java.lang.Class#getModule method in our Class model returns the correct (non-null) module for that class.

PR #136, opened 2018-08-01
Add method java.lang.String.getBytes(byte[], int, byte)

### LINKED ISSUE #134: NoSuchMethodError: java.lang.String.getBytes([BIB)V
Runtime error "java.lang.NoSuchMethodError: java.lang.String.getBytes([BIB)V' when running gov.nasa.jpf.test.java.lang.StringTest

    [junit] java.lang.NoSuchMethodError: java.lang.String.getBytes([BIB)V
    [junit]     at java.lang.AbstractStringBuilder.putStringAt(AbstractStringBuilder.java:1641)
    [junit]     at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:513)
    [junit]     at java.lang.StringBuffer.append(StringBuffer.java:275)
    [junit]     at gov.nasa.jpf.test.java.lang.StringTest.testConstructors(StringTest.java:195)
    [junit]     at java.lang.reflect.Method.invoke(gov.nasa.jpf.vm.JPF_java_lang_reflect_Method)
    [junit]     at gov.nasa.jpf.util.test.TestJPF.runTestMethod(TestJPF.java:648)

gayanW:
Add method java.lang.String.getBytes(byte[], int, byte) to MJI model class for java.lang.String

This also fixes #134:
```
[junit] java.lang.NoSuchMethodError: java.lang.String.getBytes([BIB)V
[junit] at java.lang.AbstractStringBuilder.putStringAt(AbstractStringBuilder.java:1641)
```

### DIFF
--- a/src/classes/modules/java.base/java/lang/String.java
+++ b/src/classes/modules/java.base/java/lang/String.java
@@ -241,6 +241,19 @@ implements java.io.Serializable, Comparable<String>, CharSequence {
 		return StringCoding.encode(x, coder(), value);
 	}
 
+	/**
+	 * If two coders are different and target is big enough,
+	 * invoker guarantees that the target is in UTF16
+	 */
+	void getBytes(byte dst[], int dstBegin, byte coder) {
+		assert coder == UTF16;
+		if (coder() == coder) {
+			System.arraycopy(value, 0, dst, dstBegin << coder, value.length);
+		} else { // this.coder == LATIN1 && coder == UTF16
+			StringLatin1.inflate(value, 0, dst, dstBegin, value.length);
+		}
+	}
+
 	native public byte[] getBytes();
 	@Override
 	native public boolean equals(Object anObject);


src/classes/modules/java.base/java/lang/String.java
```diff
@@ -241,6 +241,14 @@ public char charAt(int index) {
 		return StringCoding.encode(x, coder(), value);
 	}
 
+	void getBytes(byte dst[], int dstBegin, byte coder) {
+		if (coder() == coder) {
+			System.arraycopy(value, 0, dst, dstBegin << coder, value.length);
+		} else {
+			StringLatin1.inflate(value, 0, dst, dstBegin, value.length);
```
cyrille-artho: This seems to cover two cases:

1. Both encodings (source, target) are the same.
2. The target encoding is UTF-8 but the source is Latin-1.

However, what about the case that the source encoding is UTF-8 and the target is Latin-1?

gayanW: The official doc comment include this:

> Invoker guarantees: dst is in UTF16 (inflate itself for asb), if two coders are different, and dst is big enough (range check)

The method is also package-private. Anyway I could add an assertion that checks whether the target is in UTF16.

cyrille-artho:
Yes, please add an assertion check and put your comment as a comment in the code.

gayanW:
Please see the made changes.

Issue #141, opened 2018-08-09
Invalid implementation in the NativePeer method for java.lang.String#concat


gayanW:
The implementation of the NativePeer method for java.lang.String#concat, is as follows:

https://github.com/javapathfinder/jpf-core/blob/a95fd23e014d39fb01754125820cbe678dd559ab/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java#L317-L324

Here the values of the `thisStr` and `otherStr` will be the same. When constructing `otherStr`, `strRef` should be used instead of `objRef`.

PR #142, opened 2018-08-09
Fix invalid implementation of NativePeer method for String#concat

gayanW:
When constructing the JVM string object `otherStr` , `strRef` should be used instead of `objRef`.

https://github.com/javapathfinder/jpf-core/blob/a95fd23e014d39fb01754125820cbe678dd559ab/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java#L317-L324

Fixes: #141


cyrille-artho:
Is there a test that catches this? I merged the other PR but in hindsight, it would be good to have a unit test here (if there is not one already).

gayanW:
> Is there a test that catches this? I merged the other PR but in hindsight, it would be good to have a unit test here (if there is not one already).

Please try running `gov.nasa.jpf.test.java.lang.StringTest#testEquals`

Issue #140, opened 2018-08-09
Invalid implementation in the NativePeer method for java.lang.String#equals


gayanW:
The implementation of the NativePeer method for java.lang.String#equals, is as follows.

https://github.com/javapathfinder/jpf-core/blob/a95fd23e014d39fb01754125820cbe678dd559ab/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java#L166-L173

Here the values of the `s2` and `s1` will be the same. When constructing `s2`,  `argRef` should be used instead of `objRef`.

PR #143, opened 2018-08-09
Fix invalid implementation of NativePeer method for String#equals

gayanW:
When constructing the JVM string object 's2' in
gov.nasa.jpf.vm.JPF_java_lang_String#equals__Ljava_lang_Object_2__Z,
`argRef` should be used instead of `objRef`.

https://github.com/javapathfinder/jpf-core/blob/a95fd23e014d39fb01754125820cbe678dd559ab/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java#L166-L173

Fixes: #140

### DIFF
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
@@ -170,7 +170,7 @@ public class JPF_java_lang_String extends NativePeer {
     }
 
     String s1 = env.getStringObject(objRef);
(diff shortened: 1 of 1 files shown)


cyrille-artho:
Is there a test that catches this?

gayanW:
The following assertion fails before the change:

https://github.com/javapathfinder/jpf-core/blob/d595544a88bad44ea2ddedbc27e3e664a5b920fd/src/tests/gov/nasa/jpf/test/java/lang/StringTest.java#L280

PR #144, opened 2018-08-15
Remove unnecessary assertion in String#getBytes(byte[], int, byte)

gayanW:
https://github.com/javapathfinder/jpf-core/blob/9ebb98cb53b952ea3d7c7dc06aa7b822ee696395/src/classes/modules/java.base/java/lang/String.java#L249-L260

Our initial conclusion that the coder is always UTF16 is wrong. As it states "If two coders are **different** and the target is big enough, invoker guarantees that the target is in UTF16 ". So the invoker guarantees not to call when the source encoding is UTF-8 and the target is in LATIN-1.

But what if the source and target encoding are in "UTF-8". So the assertion statement needs to be removed.

This also fixes:

    [junit] java.lang.AssertionError
    [junit]     at java.lang.String.getBytes(String.java:254)

The error can be seen on ObjectStreamTest, ClassLoaderTest, StringTest and in a few other tests.

### DIFF
--- a/src/classes/modules/java.base/java/lang/String.java
+++ b/src/classes/modules/java.base/java/lang/String.java
@@ -251,10 +251,10 @@ implements java.io.Serializable, Comparable<String>, CharSequence {
 	 * invoker guarantees that the target is in UTF16
 	 */
 	void getBytes(byte dst[], int dstBegin, byte coder) {
-		assert coder == UTF16;
 		if (coder() == coder) {
 			System.arraycopy(value, 0, dst, dstBegin << coder, value.length);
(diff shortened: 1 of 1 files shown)


cyrille-artho:
Thanks for catching this. In this case, please turn the comment in the "else" condition into an assertion.

gayanW:
Done. 

Please delete the [java-10_134-1](https://github.com/javapathfinder/jpf-core/tree/java-10_134-1) branch once this is being merged. The PR is initially created from a branch in my fork. But the PR did not reflect the changes when I pushed changes. So I tried pushing directly to this repo and it resulted in a new branch being created in the javapathfinder repo.

PR #146, opened 2018-08-27
Run script for Benchexec

peterschrammel:
Benchexec is the benchmarking framework used
by SV-COMP.
https://github.com/sosy-lab/benchexec (PR#309 + PR#313)

It can be run using the benchmark definition in
https://github.com/sosy-lab/sv-comp (PR#83)
and the benchmarks in
https://github.com/sosy-lab/sv-benchmarks (PR#565)

This first version of the script does the job for now.
Comments how to generalize it in future are highly appreciated.

### DIFF
--- a/bin/jpf-sv-comp
+++ b/bin/jpf-sv-comp
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# This script allows JPF to run on Benchexec.
+# https://github.com/sosy-lab/benchexec
+
+# create site.properties
+SITE_PROPERTIES=site.properties
+echo "jpf-core = `pwd`/jpf-core" > $SITE_PROPERTIES
+echo "jpf-symbc = `pwd`/jpf-symbc" >> $SITE_PROPERTIES
+echo "extensions=\${jpf-core},\${jpf-symbc}" >> $SITE_PROPERTIES
+
+# parse arguments
+BENCHMARK=""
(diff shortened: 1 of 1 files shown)


cyrille-artho:
Thank you for providing this script. We are always happy about new contributions.
Do you have an example that shows how to run the script with a target?
Are there examples that do NOT require jpf-symbc? If all uses require jpf-symbc, then it may be better to put that script there. Otherwise, if there are examples with only jpf-core, please provide one. It may also be good to add a Wiki page describing the script. If you'd like to write one, please let us know, so we can give you write access (as that's the only way to allow edits to the wiki).

peterschrammel:
> Do you have an example that shows how to run the script with a target?

I will add a description that can then be added to the wiki too.

> Are there examples that do NOT require jpf-symbc?

I haven't tried that yet, but I'll give it a go.

> If all uses require jpf-symbc, then it may be better to put that script there. 

Can you point me to the 'current' jpf-symbc repository please?

cyrille-artho:
There is a mirror of the internal source repository of jpf-symbc at
https://github.com/SymbolicPathFinder/jpf-symbc/
It would be great to see if we can use sv-comp-run-script on jpf-core as well, as that would allow us to host it here and also use it to benchmark various variations of search heuristics or other parameter settings in jpf-core.

Issue #151, opened 2018-08-30
Arrays.get/setShort does not work


jtoman:
Nor does reflectively setting or getting a short field. MJIEnv (erroneously) forward get/setShortField to get/setIntField, which fails with a type mismatch error.

cyrille-artho:
A (failing) unit test for that would be great.

jtoman:
Hi @cyrille-artho this issues (and all the others I opened) will have PR's coming shortly which will include the unit tests.

PR #155, opened 2018-08-31
Fix get/setShort for field and array reflection

jtoman:
Fixes and unit tests for #151. As mentioned in the issue, env.get/setShortField was incorrectly forwarding to get/setIntField.


cyrille-artho:
Many thanks!

Issue #147, opened 2018-08-30
Capturing null-valued free variables throws an exception


jtoman:
At lambda creation time, if one of the variables captured is nulll, then setFuncObjFields fails. This just needs a null check to fix.

PR #156, opened 2018-08-31
Fixes null captured parameters

jtoman:
Fixes and unit test for #147. As mentioned in the issue, there was no null check when unpacking object values.


jtoman:
This test failed because of some pom error. I don't know how to retrigger a build.

PR #157, opened 2018-08-31
Fixes setCause on default exception

### LINKED ISSUE #148: Cannot set new cause on an exception that uses no-arg throwable constructor
You cannot do `new RuntimeException().setCause(...)`. The cause field is not set to this in the default throwable constructor.

jtoman:
Fixes #148. We just need to set cause in the no-arg throwable constructor to this.

### DIFF
--- a/src/classes/java/lang/Throwable.java
+++ b/src/classes/java/lang/Throwable.java
@@ -44,7 +44,7 @@ public class Throwable {
     } catch (ClassNotFoundException e) {
       throw new NoClassDefFoundError("java.lang.StackTraceElement");
     }
-     
+    cause = this;
     fillInStackTrace();
   }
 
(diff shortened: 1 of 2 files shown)


cyrille-artho:
This patch seems to break the behavior that the default (uninitialized) cause of a `Throwable` is `null`. The test also has no assertion (it would be good to check the result of `getCause`).
I don't have access to a computer with Java 8 right now so I can't run the example.

cyrille-artho:
I see. However, using "this" as a sentinel complicates the code and seems unnecessary.
How about changing the checks against "this" to checks against null, and adapting the code accordingly? This will make the code change a bit bigger, but if this works (against all existing unit tests), it will make the code clearer and closer to the API specification.
Also, the unit test should check if getCause() returns the exception that was passed to initCause(), i.e., NPE.

jtoman:
Sure but that design decision predates this patch and the setting of cause
to this is found in the other constructors. This PR just brings the noarg
constructor in line with the rest of the class. Refactoring is a separate
effort that falls outside the scope of this fix.

cyrille-artho:
I see, if other constructors rely on this "quirky" implementation, then it's best to leave that for now.
The rationale behind that implementation is not clear to me, but as you said, it's been in the code for a long time.

PR #159, opened 2018-08-31
Fixes default method resolution

### LINKED ISSUE #154: Default method resolution is too restrictive
The current resolution process fails if it finds two potential default methods. However, technically this resolution process should only fail if there is no most specific resolution. In other words, if there are two default methods from interfaces I1 and I2, and I2 <: I1, then we can safely ignore the default method in I1.

jtoman:
Fixes #154. Now when multiple potential default method resolutions, we check whether one type is a subinterface of the other and select the more specific type. If the two types are not in a subtyping relation, we fail as before.

In addition, the looping algorithm was changed because the old version did not consider super interfaces. This commit includes a unit test to test this behavior.


cyrille-artho:
The new unit test fails on Travis.

cyrille-artho:
Thanks, this looks non-trivial so it is good to have unit tests that verify the behavior on some examples.

PR #161, opened 2018-09-05
Improves annotation support

### LINKED ISSUE #150: Improve annotation support
The current annotation system of JPF lacks the following features:

* Annotation-valued attributes are deeply broken
* Default value reflection isn't supported
* toString() does not support primitive fields or primitive arrays beside int, long, and boolean
* Attempting to access an annotation with primitive attributes that are not int, long, or boolean will not work. Ditto for arrays thereof
* Mising annotation definitions completely tank class loading (annotations with missing types should just be skipped)
* equals() and hashCode() do not work as expected

jtoman:
Closes #150. Improves toString to cover all primitive types and annotation
attributes. Also adds support for:

* Default value reflection
* hashCode/equals on annotations
* nested annotations
* Ignoring missing annotations (which is what the JVM does)

### DIFF
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
@@ -17,16 +17,10 @@
  */
 package gov.nasa.jpf.vm;
 
-import gov.nasa.jpf.annotation.MJI;
-import gov.nasa.jpf.vm.CharArrayFields;
-import gov.nasa.jpf.vm.ElementInfo;
-import gov.nasa.jpf.vm.Fields;
-import gov.nasa.jpf.vm.Heap;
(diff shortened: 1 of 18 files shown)


src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
```diff
@@ -572,7 +570,7 @@ public int toCharArray_____3C (MJIEnv env, int objref) {
   }
 
   @MJI
-  public int format__Ljava_lang_String_2_3Ljava_lang_Object_2__Ljava_lang_String_2 (MJIEnv env, int clsObjRef, int fmtRef, int argRef) {
```
jtoman: Oops, this is from a different change

cyrille-artho:
Great work, and indeed the diff was so long that it was not shown, and I had not noticed.

PR #166, opened 2018-12-05
[fix-bug]: checking attrs is not null in method StackFrame:replaceLocalAttr( )

ngocpq:
The condition "attrs is not null" is written as (attrs==null), instead of (attrs != null).
My patch is as follow:
. -    if (attrs == null){
. +   if (attrs != null){
.          attrs[index] = ObjectList.replace(attrs[index], oldAttr, newAttr);    
.       }


ngocpq:
Hi, 
I have add a test case that reveal the bug in the last commit.

cyrille-artho:
Great, thank you very much!

ngocpq:
My pleasure.

PR #170, opened 2019-01-03
BitSet1024Test.java

durubayram:
Expected errors were hidden.


cyrille-artho:
Thanks for this PR.
Can you please make the PR against the master branch? We currently don't develop on the "java-8" branch, and will merge "master" into it once JPF is ready for Java 11.
It is a good idea to merge "master" into "java-8" periodically as we're getting close to this point, but I will do that without adding new commits.

Therefore, please make a new PR against "master" instead of "java-8", so only one commit will appear.

PR #171, opened 2019-01-13
Issue #163: Support java.nio.Buffer

krizhan:
Hey,
Sorry for the delay as I was busy with school assignments. Anyways I made some changes to the classes and no longer get java.lang.NoSuchMethodException and the jmbc regression test runs for me. The errors I get from running the test are expected (such as no heap space and NoSuchElementException). Let me know if there are any issues and concerns with the code and I will gladly address it. 

-Krizhan

### DIFF
--- a/src/classes/java/nio/Buffer.java
+++ b/src/classes/java/nio/Buffer.java
@@ -18,9 +18,26 @@
 package java.nio;
 
 public abstract class Buffer {
+
 	protected int position;
 	protected int capacity;
 	protected int limit;
+	protected int mark = -1;
+
+
+	// Todo: There exists other missing methods in this class
+	// which may throw NoSuchMethodException errors to users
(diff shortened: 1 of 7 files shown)


cyrille-artho:
Can you please add a unit test that requires this patch in order to pass?
Cross-posting this from the issue tracker:

For the test, have you tried this:
Character.forDigit(digit, radix);
Maybe this is enough to make old JPF (without your patch) fail.
If not, please try both cases from the test in https://github.com/sosy-lab/sv-benchmarks/blob/master/java/jbmc-regression/StaticCharMethods05/Main.java, as a unit test.
It is not necessary to have a non-deterministic choice about the operation; it's also OK to execute both blocks inside the switch block, sequentially.

krizhan:
Ohh okay I understand. Thanks for clearing it up.

PR #172, opened 2019-01-27
Rectifies the java.lang.NoSuchMethodError when usinf printf

manish3499:
The model and peer classes java.util.regex.Matcher are modified to accommodate the overloading of the function find() , the lack of which was causing the error.


cyrille-artho:
Thank you for fixing this!

Issue #180, opened 2019-03-23
Incorrect behavior of Calendar-related native peers


fresheed:
Playing around with JPF I have discovered a problem which boils down to following:

```
        Calendar cal1=Calendar.getInstance();
        cal1.set(Calendar.HOUR_OF_DAY, 10);
        Assert.assertEquals(10, cal1.get(Calendar.HOUR_OF_DAY)); // ok

        Calendar cal2=Calendar.getInstance();
        cal2.set(Calendar.YEAR, 2014); // year is important
        cal2.set(Calendar.HOUR_OF_DAY, 10); // value is not important
        Assert.assertEquals(2014, cal2.get(Calendar.YEAR)); // ok
        Assert.assertEquals(10, cal2.get(Calendar.HOUR_OF_DAY)); // on JPF VM gets 11 instead

```
So, setting hour along with year somehow shifts hour forward by one. It doesn't depend neither on sets order nor on hour value. However, it depends on year value (with 2015 it runs correctly, but with 2013 fails too) and concrete parameter being set (Calendar.MONTH change is processed correctly).
There are native peers for Calendar, Locale, Timezone and few other classes related to date processing, so the bug should be related to them. 
I haven't managed to locate concrete problem source yet, mostly because I haven't discovered a convenient way to debug apps being run on JPF VM. There is jpf-inspector project - should I use it? Or are there other ways to debug JPF native peers?

cyrille-artho:
This is an interesting and strange bug, because the native peer class in jpf-core implements a very minimal functionality for Calendar: src/peers/gov/nasa/jpf/vm/JPF_java_util_Calendar.java has only one method. Other methods are delegated to java.util.Calendar on the host JVM.

Which Java version are you using? How does your program behave when run normally on Java?

It is indeed not easy to debug this; however, you can make a unit test, so we can at least try this test on different versions of Java (the host JVM).

fresheed:
Expected behavior is observed on HotSpot 1.8.0_201.
I suspect the problem not in Calendar peer itself, but rather in Locale peer which is called by Calendar, namely in getLocale method.
Do you mean a pull request with (commented out) unit test?

cyrille-artho:
If you add a test that is enabled, Travis will run it, and we can see the outcome there.
I will then also be able to check the outcome locally (by downloading your pull request).
So please make a pull request with the test enabled.
If it fails, we can still comment out the line that fails.

cyrille-artho:
[...]

PR #181, opened 2019-03-27
Calendar peer bug demonstration

fresheed:
Demonstration of a bug described in #180 . To observe it, run this class on JPF.


fresheed:
Sorry, my bad. Didn't notice an instruction at wiki on writing JPF tests. I have updated it accordingly and now the bug can be observed by running `./bin/test gov.nasa.jpf.test.java.util.CalendarTest`

cyrille-artho:
Thank you for the test. Interestingly, it passes on the system run by Travis. I will merge it and try it on my own system.
It could be a bug on the host JVM setup or locale... in any case, it is useful to have a test for that.

cyrille-artho:
The test works on my computer. Perhaps it is configuration-dependent. In that case, it would be useful to find out (some day) what configuration aspect causes this test to fail on some systems.

PR #182, opened 2019-03-27
Add Peer for java.util.concurrent.atomic.AtomicReference

Jentsch:
Hi needed this to get a minimal Scala-Futures example running. Maybe this is useful for others too. The code is basically the same as for AtomicInteger.

### DIFF
--- a/src/main/gov/nasa/jpf/vm/NativePeer.java
+++ b/src/main/gov/nasa/jpf/vm/NativePeer.java
@@ -35,7 +35,7 @@ import java.util.logging.Level;
 /**
  * native peer classes are part of MJI and contain the code that is
  * executed by the host VM (i.e. outside the state-tracked JPF VM). Each
- * class executed by JPF that has native mehods must have a native peer class
+ * class executed by JPF that has native methods must have a native peer class
  * (which is looked up and associated at class loadtime)
  */
 public class NativePeer implements Cloneable {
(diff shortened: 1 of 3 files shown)


cyrille-artho:
Looks good! Is it possible to add a small unit test? (Perhaps you can create a unit test in Java to avoid a dependency on Scala for only a test?)

Jentsch:
Hi,
added a unit test for `getAndSet' and 'compareAndSet'. I hope it's good enough.

I didn't implemented the `weak*` and `lazy*` methods, since I don't know how to handle implement them correctly. The implementation of `AtomicInteger` `lazySet` and `weakCompareAndSet` doesn't seems to be right, because they behave like there non lazy and weak counterparts.
All other methods of `AtomicReference` should be implemented in terms of `getAndSet` and `compareAndSet` or directly modify the `value` field.

The only additional code to get a minimal example with Scala-Futures running is a very simple `ExecutionContext` (Scalas version of a thread pool) that simply spawns threads with out pooling them. [You can find it here](https://github.com/Jentsch/modelchecker/blob/master/jpf/src/main/scala/test/JpfExecutionContext.scala)

cyrille-artho:
Thanks for the tests. It is fine to have an incomplete model class, and usually better than stubs that do not behave correctly. If needed later, stubs can still be added and improved. (If stubs are added without being used anywhere, and without test cases, then there is a bigger risk that they behave incorrectly. We have found a couple of cases with the SV-COMP-2019 tests.)

PR #184, opened 2019-03-31
Updated Search.java documentation

Hayawi:
Fixed spelling and grammar errors, added verbosity to existing javadoc
and added javadoc where none previously existed.

### DIFF
--- a/src/main/gov/nasa/jpf/search/Search.java
+++ b/src/main/gov/nasa/jpf/search/Search.java
@@ -1051,3 +1051,4 @@ public abstract class Search {
     return true;
   }
 }
+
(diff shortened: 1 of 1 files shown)


cyrille-artho:
Thanks for the update. The current version removes the newline character at the end of the file; however, it is customary to keep that, for various reasons (usage of grep, sed, diff, patch, etc.).
Please add the newline character at the end back.

Hayawi:
@cyrille-artho My mistake, I've committed an updated Search.java with the newline character added back in. Please let me know if it's up to par.

cyrille-artho:
Thanks, looks good! Could you please make a few more small changes:
1.
`to the {@code properties} ArrayList`
->
`to {@code properties}`
The properties are implemented as an ArrayList for now, but there is no intrinsic need for them to be a list (they could be a set or a subclass of a Properties class). It's better not to have too much redundancy in the comments.

2.
` {@code error(Property property, Path path, ThreadList threadList)}`
->
`{@code error}`
Again, if the method signature changes, the comment will be outdated and wrong; the easiest way to avoid this is by keeping the comment simple.

Other than that, I could not find any place with possibly redundant information, so I think we can merge your PR after these changes, unless someone else notices something.

Hayawi:
Looked over the documentation and made changes to what you suggested as well as a couple more sections which exhibited the same problems. It should contain less redundancy now.

cyrille-artho:
Very nice, thanks! Merging it now.

Issue #187, opened 2019-05-04
Fix Travis setup


cyrille-artho:
It looks like Travis no longer supports Java 8. Recent commits no longer build. Someone should check if/how we can run the current version of JPF on Travis.

manish3499:
Hi @cyrille-artho , I'll look into it.

PR #189, opened 2019-05-19
Fix issue #187: use ubuntu trusty version for travis build environment

yuvaraj-anbarasan:
There's a known issue in travis xenial environment. It only supports JDK versions 9-11.


cyrille-artho:
Great, this allows us to keep using oraclejdk8 for the time being.

PR #190, opened 2019-05-25
upgrade to gradle 5.4.1

yuvaraj-anbarasan:


cyrille-artho:
Thanks! It seems that there are no deprecation warnings about older features carried over from Gradle 4, at least not in the Travis log. So we can use this version.

Issue #197, opened 2019-06-26
A bug in java.lang.String.lastIndexOf


rtrimana:
There seems to be in issue with this native method in JPF_java_lang_String.java.
Found this bug when trying to use the BigInteger library with the following initialization:

`BigInteger bi = new BigInteger("-1");`

In the BigInteger constructor, the JPF's lastIndexOf() incorrectly returns -1 instead of 0 for:

`String val = "-1";
int index1 = val.lastIndexOf('-');`

This is caused by the following loop in the native method implementation of lastIndexOf() in JPF_java_lang_String.java.
Instead of: 

`for (int i = fromIndex; i > 0; i--) {
      if (values[i] == c) { return i; }
    }`

, it should have been ('>=' instead of '>'):

`for (int i = fromIndex; i >= 0; i--) {
      if (values[i] == c) { return i; }
    }`

This is the full method implementation:

    public int lastIndexOf__II__I (MJIEnv env, int objref, int c, int fromIndex) {
        int vref = env.getReferenceField(objref, "value");
        ElementInfo ei = env.getElementInfo(vref);
        char[] values = ((CharArrayFields) ei.getFields()).asCharArray();

        int len = values.length;

        if (fromIndex < 0) { return -1; }
        if (fromIndex > len - 1) {
            fromIndex = len - 1;
        }

        for (int i = fromIndex; i >= 0; i--) {
            if (values[i] == c) { return i; }
        }

        return -1;
    }

cyrille-artho:
Thank you for this bug report; this is indeed a bug, and your analysis looks correct.
Would you like to create a unit test and fix (patch) for this? We can help you with that.
The best idea is to look at existing unit tests and then add a new one in this way:

```
@Test public void test... () {
  if (verifyNoPropertyViolation(){
    ... code to verify by JPF
  }
}
```

rtrimana:
I would be happy to try that. Could you please guide me further?
I think I need to follow the pointers [here.](https://github.com/javapathfinder/jpf-core/wiki/Writing-JPF-tests)
And then I need to add that unit test [here.](https://github.com/javapathfinder/jpf-core/blob/master/src/tests/gov/nasa/jpf/test/java/lang/StringTest.java) 
Am I correct?

cyrille-artho:
Yes, please add a unit test there.
You can change your one-liner to use an assertion:
```
String val = "-1";
assert(val.lastIndexOf('-') == 0); // test against off-by-one error (issue #197)
[...]

PR #200, opened 2019-06-27
Fixed off-by-one error in String.lastIndexOf() method (#197)

rtrimana:
Fixed an off-by-one error in the model class implementation of the method `String.lastIndexOf()` in `JPF_java_lang_String`. Initially the for-loop counter always excludes 0 as the last index, whereas it should have been included in the search.


rtrimana:
That's great! But I am confused. I thought this was a pull request that you asked? :)
This is the link to it: https://github.com/javapathfinder/jpf-core/pull/200
Did I miss anything?

cyrille-artho:
Thank you, merged!

Issue #201, opened 2019-07-03
A bug in java.lang.StringBuilder.append(char c)


rtrimana:
I have just observed a different behavior in the method `append(char c)` compared to the standard java.lang.StringBuilder library for the following code:

	  StringBuilder sb = new StringBuilder(0);
	  sb.append('[');
	  sb.append(']');
	  System.out.println(sb.toString());

On standard JVM:
`[]`

On JPF:
```
---------------------- JPF error stack trace ---------------------
gov.nasa.jpf.JPFException: illegal array offset: 0
	at gov.nasa.jpf.vm.ElementInfo.checkArray(ElementInfo.java:1180)
	at gov.nasa.jpf.vm.ElementInfo.setCharElement(ElementInfo.java:1279)
	at gov.nasa.jpf.vm.MJIEnv.setCharArrayElement(MJIEnv.java:408)
	at gov.nasa.jpf.vm.JPF_java_lang_StringBuilder.append__C__Ljava_lang_StringBuilder_2(JPF_java_lang_StringBuilder.java:154)
```

I am proposing the following fix in the native method:
```
public int append__C__Ljava_lang_StringBuilder_2 (MJIEnv env, int objref, char c) {
    int aref = env.getReferenceField(objref, "value");
    int alen = env.getArrayLength(aref);
    int count = env.getIntField(objref, "count");
    int i;
    int n = count +1;

    // TODO: Need to allocate something if the initial array is really empty
    if (alen == 0) {
      $init____V (env, objref);
      aref = env.getReferenceField(objref, "value");
      alen = env.getArrayLength(aref);
    }
    if (n < alen) {
      env.setCharArrayElement(aref, count, c);
    } else {
      int m = 3 * alen / 2;
      int arefNew = env.newCharArray(m);
      for (i = 0; i < count; i++) {
        env.setCharArrayElement(arefNew, i, env.getCharArrayElement(aref, i));
      }
      env.setCharArrayElement(arefNew, count, c);
      env.setReferenceField(objref, "value", arefNew);
    }
    
    env.setIntField(objref, "count", n);
    
    return objref;
    
  }
```
Basically `alen` needs to be checked and if it is 0, we need to allocate some array to allow appending new characters.

cyrille-artho:
Thank you for spotting and diagnosing this. Please make a patch with the example as a unit test and the fix for the native peer.
[...]

PR #202, opened 2019-07-04
Fixed zero-capacity error in StringBuilder.append(char c) method (#201)

rtrimana:
Fixed a zero-capacity error in the model class implementation of the method `StringBuilder.append(char c)` in `JPF_java_lang_StringBuilder`. Initially when capacity is set to 0 for the constructor during a StringBuilder object instantiation, the JPF's implementation for the `append(char c)` will throw an `illegal array offset: 0` exception when the method is called. The behavior for the standard StringBuilder library is to allow such appends even when the capacity has been first set to 0. Thus, for the JPF's implementation, we first need to allocate a default array if the initial capacity is 0.


cyrille-artho:
Thanks!

Issue #196, opened 2019-06-25
JPF cannot find model classes with module system on Java 11


cyrille-artho:
Up until Java 8, model classes were compiled to `build/classes`. For example, `java.lang.Thread` was under `build/classes/java/lang/Thread.class`, and the same relative path in `jpf-classes.jar`.

With the module system, the path changes to `build/classes/modules`, so `Thread` would be under `build/classes/modules/java.base/java/lang/Thread.class`.

In `ClassPath.java`, `findMatch` (line 103) is already invoked first with `java/lang/Thread.class`, then with `java.base/java/lang/Thread.class`.

Perhaps it is possible to modify the code such that the added `java.base` becomes `modules/java.base`. Otherwise, we can intercept each call in `ClassPath.java` before the loop (line 104), and

1. inspect the first element in the package name.
2. if it starts with "java.", then prepend "modules".
3. continue loading the class (inside the loop) as before.

This should eliminate issues with JPF model classes not being found. At the moment, JPF cannot find `Thread$Permit`, which only exists in the JPF model class.

cyrille-artho:
Resolved as part of PR #210.

PR #206, opened 2019-07-10
Fixed issue #196: added modules to classEntryURL

yuvaraj-anbarasan:


cyrille-artho:
Thanks. As pointed out by Franck, the URLs should have "/" instead of File.separator.
Please update the PR accordingly.

PR #207, opened 2019-07-15
Fixed issue #196: Added modules to classEntryURL

yuvaraj-anbarasan:


cyrille-artho:
Does this fix improve the outcome of certain unit tests?

yuvaraj-anbarasan:
This fixes the StackOverflow error which was occurring earlier. But another
way to fix this is to change the destination dir of module classes to
build/classes
instead of build/classes/modules and this works.

PR #211, opened 2019-09-18
Missing Thread id (tid) bug fix during jpf running

jsrudani:


cyrille-artho:
Hi,
Thank you for the PR.
The last commit reverts to an earlier one and removes any examples/tests etc. that may have been part of the PR otherwise. The current PR has only one field declaration.

There is no need to declare the tid in the model class, as it's handled by the native peer (which uses ThreadInfo for this).

PR #212, opened 2019-11-07
fixed random search where only one path is explored even with limit > 1

richardnnn:
Right now the `RandomSearch` always returns after finishing one path, this PR tries to fix this problem and allow it to explore mutiple paths specified by `search.RandomSearch.path_limit`. The previous code will fail the added test but should pass after merged. 

Now after `cg.randomize_choices` set to not `NONE` , the search will randomly try different paths.

### DIFF
--- a/src/main/gov/nasa/jpf/search/RandomSearch.java
+++ b/src/main/gov/nasa/jpf/search/RandomSearch.java
@@ -29,7 +29,8 @@ import gov.nasa.jpf.vm.RestorableVMState;
  * going forward() until there is no next state then it restarts the search 
  * until it hits a certain number of paths executed
  *
- * <2do> this needs to be updated & tested
+ * To explore different paths, cg.randomize_choices shouldn't be set to NONE
+ *
  */
 public class RandomSearch extends Search {
   int path_limit = 0;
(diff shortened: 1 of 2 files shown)


ankushdesai:
@cyrille-artho This pull request fixes a bug in RandomSearch implementation.

src/main/gov/nasa/jpf/search/RandomSearch.java
```diff
@@ -66,12 +67,12 @@ public void search () {
           }
         }
 
-        if (isEndState()){
-          return;
-        }
+//        if (isEndState()){
+//          return;
+//        }
```
cyrille-artho: Remove code rather than commenting it out.

src/main/gov/nasa/jpf/search/RandomSearch.java
```diff
@@ -66,12 +67,12 @@ public void search () {
...
 
         depth++;
-
+        //System.out.println("path " + paths + ", depth " + depth);
```
cyrille-artho: remove

src/main/gov/nasa/jpf/search/RandomSearch.java
```diff
@@ -85,6 +86,7 @@ public void search () {
         depth = 1;
         vm.restoreState(init_state);
         vm.resetNextCG();
+
```
cyrille-artho: remove

cyrille-artho:
Thank you very much for finding and fixing this as well as adding a test.
If you don't mind, would you please remove commented-out code as well as the extra line break that you inserted?

cyrille-artho:
Thank you very much! I have merged the two patches into one and merged it.

richardnnn:
Happy to help :)

Issue #229, opened 2020-07-10
Use the host JVM to handle string concatenation 


amgad-rady:
1. Fixed a bug where string concatenation with more than three variables would fail.
2. Changed the method FunctionObjectFactory.makeConcatWithStrings to use the bootstrap method StringConcatFactory.makeConcatWithConstants to perform string concatenation on the host JVM.

PR #230, opened 2020-07-10
String concat debugging

amgad-rady:
Pull request related to issue #229 on tracker.


cyrille-artho:
Thanks! I have squashed the commits (into one commit), and removed commit messages from temporary code.

PR #264, opened 2020-08-22
fix #261: Check constructor access level in java.lang.reflect.Constructor.newInstance native peer

### LINKED ISSUE #261: `java.lang.reflect.Constructor.newInstance` native peer doesn't check for access level
```
class Base {
    static class Inner {
        private Inner () {
        }
    }
}
public class Test {
    public static void main(String args[]) {
        Class innerClass = Base.Inner.class;
        try {
            Object innerClsInstance = innerClass.getDeclaredConstructor().newInstance();
        } catch(Exception e) {
            System.err.println(e);
        }
    }
}
```
The above code executed on the host JVM, throws `java.lang.IllegalAccessException`, but when executed on JPF, doesn't throw any exception since java.lang.reflect.Constructor.newInstance` native peer doesn't check for access level.

yuvaraj-anbarasan:


cyrille-artho:
Looks good, thanks!

PR #271, opened 2020-10-13
Improved Docs

Varun11940:


cyrille-artho:
Thanks; it seems parentheses are not necessary for `println` in gradle. Are they perhaps needed in a newer version (so the change would facilitate a future migration)?

PR #275, opened 2020-11-03
Fix FileIOTest in Java 11.

mmuesly:


mmuesly:
I have to rethink the PR. It is not the right fix, after reading in the current API of Java 11.

mmuesly:
The second commit repairs it on a better place.

cyrille-artho:
Thanks!
The Travis setup is currently broken, but I was able to run the tests locally.
Do you know if using `\n` as a line separator is fully portable? We generally do not test much on platforms using non-standard EOL encodings, but I wonder if (for example) regular console output works properly without the MS-DOS `\r`.
Other than that, the patch looks good.

src/peers/gov/nasa/jpf/vm/JPF_java_lang_System.java
```diff
@@ -83,10 +83,15 @@ int createPrintStream (MJIEnv env, int clsObjRef){
...
   @MJI
-  public int createSystemErr____Ljava_io_PrintStream_2 (MJIEnv env, int clsObjRef){
-    return createPrintStream(env,clsObjRef);
+  public int createSystemErr____Ljava_io_PrintStream_2(MJIEnv env, int clsObjRef) {
+    return createPrintStream(env, clsObjRef);
+  }
+
+  @MJI
+  public int lineSeparator____Ljava_lang_String_2(MJIEnv env, int objRef) {
+    return env.newString(System.lineSeparator());
```
mmuesly, line 94: The peer maps the line Separator to the line Separator of the host JVM in this line of code.

cyrille-artho: Thanks, in that case the PR is good to go!

PR #293, opened 2021-08-14
Himanxu Removes some spaces

Himanxu1:


cyrille-artho:
Hi,
Thank you for your initial contribution. I agree that these two empty lines in class `Error` can be removed, but there is no need to add empty lines (or empty comments) elsewhere.
If you would like to clean up some code, please try to cover several classes (for example, in the same package) at once in a uniform way, and try to document how you do it, so we can write this as a code style guideline for the future.

Himanxu1:
Thanks, @cyrille-artho .I am new to Open Source and want to contribute to this project. I know Java But I don't know how can I use  it herr.

mamadaliev (approved):

cyrille-artho:
Hi Himanshu,
You simply need to undo your changes to the two other files and push again. This pull request will get updated, and it will contain only the changes in `Error.java`. That's the change we'd like to keep, but without the extra empty lines in the other two files.

Issue #290, opened 2021-08-04
Publish JPF to a Maven repo


iirekm:
... or at least use standard Gradle tasks (e.g. `jar` instead of `buildJars`) so https://jitpack.io/ can be used to download pre-built JARs

cyrille-artho:
There is a default task `jar`, so I'm not sure how to declare a renamed `buildJars` without interfering with that one.
We are happy for improvements in `build.gradle` but not always successful in fixing such issues ourselves. A pull request would be highly appreciated!

mmuesly:
#296 allows to use jitpick.io with jpf-core. Does this already fit for your case @iirekm?

varad64:
Hello @cyrille-artho, with @mmuesly's patch now refactored and applied to `java-10-gradle`, the functionality to publish to the local maven repository is supported and I think the issue can now be closed.

Thanks!

PR #296, opened 2021-09-07
Add ptml plugin

mmuesly:
I just have read #290. Actually, I have a version of the gradle file that allows to use jpf-core as build by jitpick.io.

See our jdart buildfile for more details: https://github.com/tudo-aqua/jdart/blob/2c3e24be0a7726e5f3821928f157a4ad72b0e5d2/build.gradle.kts#L40


cyrille-artho:
Thanks, this looks good! (Admittedly, I am not familiar with publishing to maven repos, but all CI tests except `ExchangerTest` pass.)
Thank you very much!

PR #299, opened 2021-09-18
remove the empty bodies of some methods in Verify.java, and add native qualifiers for them

y553546436:
As discussed in #297.


cyrille-artho:
Thank you, and it looks like CI finally works again like a charm!

PR #303, opened 2021-11-22
Added Unit test example 

gaurangkudale:
Added `IntStream` example to `src\tests\StreamTest`


gaurangkudale:
Hi @cyrille-artho 
I don't know why it is showing previous commits will it affect this PR?

cyrille-artho:
You probably have to pull the earlier changes from `javapathfinder:master` first. This will bring your fork up to date with the main repository.

gaurangkudale:
So unit test example is right or do I need any modifications>

cyrille-artho:
As for the test itself, it should automatically verify something (without using `print`). The easiest way is to have a local counter `i` that runs alongside the final stage of the stream, and you check at each step that the value from the stream is the same as the value of the counter (while incrementing the counter). This way, we have a simple way to test that the stream returns all numbers in sequence (if used sequentially).

Issue #302, opened 2021-11-22
Support for `java.util.stream` library


cyrille-artho:
Java 8 has introduced a new library to deal with large data structures, Java Streams:

https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html

We don't have any unit tests for the new Java streams collections (`java.util.stream`), and thus we don't know how well JPF handles them (if at all).
A few examples (unit tests) would be helpful to see if JPF can handle these cases at all (there may be missing native peer classes/methods), and what problems there might be.

gaurangkudale:
I'll look into this issue and also create some unit tests for this issue to check whether it's working or not till then can you please assign this issue to me so that I'll start working on it

gaurangkudale:
Do I only need to create unit test for the `java.util.stream` is that right?

cyrille-artho:
You can start with a test that builds a stream pipeline for one of the data structures in that package (e.g., `IntStream`) and creates a stream that does something useful.
You can adapt existing open-source code as a simple unit test as long as the license of the existing code is compatible with the license of JPF (ideally, having the Apache License v2), or you can create your own example and provide it to the JPF project under the Apache License.
The task is quite open-ended at the beginning: create a small example that runs on the normal JVM and produces a meaningful result, and then see what happens inside JPF.
Ideally, the example is less than 20 lines of code (plus import statements).

gaurangkudale:
Okay got it

gaurangkudale:
Hi @cyrille-artho 
I've added the unit test example of IntStream and created the pull request

achintya-7:
Is there requirement for more unit tests. I am just asking because this issue is still open.

cyrille-artho:
Guarang's work is in progress and close to finished. It seems that the streams library (at least for the common function) uses no native code and likely works as expected for simple cases.
This means we do *not* need additional unit tests for Java streams right now. Unit tests are very welcome for other parts of Java's libraries, especially if they help to narrow down the reason for compatibility problems with Java 11+.
So if you look at issue #274 and can find a way to write a unit test that is simpler than one of the existing tests and still shows the same type of problem, that will be very useful to find a bug fix.

achintya-7:
Sure, I'll look into it

cyrille-artho:
Addressed by PR #305

PR #305, opened 2021-11-23
Unit Test for java.util.stream

gaurangkudale:
I've created a simple unit test for `java.util.stream` in this path `src\tests\StreamTest`


gaurangkudale:
Hi @cyrille-artho,
 As you said I've changed the code where the print statement is removed and I have got a few questions

1. Should I write the test cases for all of the methods present in `java.util.stream`?
2. As I've written the unit test for `IntStream` is it working fine? and is it mergeable to the main jpf-core

gaurangkudale:
Hi@cyrille-artho 
Can you please solve my doubts so that I can keep contributing

cyrille-artho:
The version I can see right now prints all the numbers. Typically, in a unit test, we want to avoid side effects such as printing on the console, and instead verify as much as possible using assertions.
Try to rewrite your code such that it uses `assertEquals` from JUnit.
To answer your first question, it is not necessary now to test all methods of Java Streams. If we test one or two methods, we can see that streams are supported as such in JPF. If a particular functionality does not seem to work well, we can create more unit tests later.

gaurangkudale:
Hi @cyrille-artho 
I've updated the code such that it is not printing the number in the console

gaurangkudale:
Please check the `src\tests\StreamTest` folder to verify

cyrille-artho:
Hi,
Thank you for the update. The challenge in the test is to ensure that `doSomething` is called in the right order by the stream. Each number from 0 to 9 should appear once in succession. Probably it's necessary to create a new (inner) class with a field that remembers the value of the counter, so it can count independently from 0 to 9 and verify that the call sequence is correct. Having a redundant counter allows you to encode the property in an assertion:
`assertEquals(value == expected);`
This assumes that `doSomething` is called with a parameter (`value`) and that `expected` is the name of the counter that goes from 0 to 9.
Also, please note the typo in `doSomething` at the moment, which you can fix alongside the suggested rewrite.

gaurangkudale:
Hi @cyrille-artho Currently I've started learning JUnit for testing so can you please tell me what exactly I should do with `doSomething`?

cyrille-artho:
A useful unit test has at least one assertion (such as `assertEquals`) that checks if the value computed by a given function matches the expected value. In your example, the expected values for each call are a series from 0 to 9.
You therefore need to create a small class that allows you to instantiate an object that counts every time `doSomething` is called. Inside `doSomething`, you use your own counter and compare it to the value returned by the stream.
This way, you can confirm that the stream returns the numbers in the same order (if it is a sequential stream).

gaurangkudale:
Hi @cyrille-artho Is there something wrong about Unit Test

gaurangkudale:
Hi @cyrille-artho 
I've done some changes in the code but it is running fine on my machine why it is not passing the test cases over there any specific reason behind that?
Link of screenShots 1 :-  [https://kjeieduin-my.sharepoint.com/:i:/g/personal/14seit_tcoer_kjei_edu_in/EfzaXDUtyIpCsBz0xt5Mg-EBizf7AwOg9Y4XvBhGhlx8jw?e=f0KjoR](url)

Link 2  [https://kjeieduin-my.sharepoint.com/:i:/g/personal/14seit_tcoer_kjei_edu_in/EXQDEArP2cVLi7a-rY5WbDABXQ3-VhVm1L9YTjV3Lg06og?e=0cNmnt](url)

gaurangkudale:
Is there anything I can do?

gaurangkudale:
Hi @cyrille-artho 
Can you check whether it is working on your system?

cyrille-artho:
It looks like you import the assertions from the wrong package. Look at some existing tests to see which packages to import. Maybe it will suffice if you extend class `TestJPF`, as that is necessary for tests that run inside JPF; methods like `assert` are inherited then.

Also, move the test classes to their own package, e.g., to, `src/tests/java8` as package `java8`, to avoid cluttering the top-level directory with test files.

gaurangkudale:
Hi @cyrille-artho Did the changes as you suggested now can you please review the code and please let me know if  there are any changes

gaurangkudale:
HI @cyrille-artho I think the changes are working fine can you please review the code

gaurangkudale:
Hi @cyrille-artho  I think the changes are working fine can you please review the code

cyrille-artho:
Thanks for the update. I see that the code now compiles and runs, and the test passes.
However, please check the content of the test. It may be better to have something more simple such as "isEven" so we don't have to worry about the recursion. Most importantly, the assertion should check something that the stream computation returns. `assertEquals(2,2);` will always succeed no matter what.

src/classes/java/lang/ThreadLocal.java
```diff
@@ -105,15 +105,13 @@ public void remove(){
     class threadSafeFormatter{
       public static ThreadLocal<SimpleDateFormat> df = ThreadLocal.withInitial(() 
           -> new SimpleDateFormat("yyyy-MM-dd"));
-  }
-  
   public class App  {
       public static void main(String[] args) throws Exception {
           threadSafeFormatter tf = new threadSafeFormatter();
           Thread t1 = new Thread();
           t1.start();
```
ramawatraksha75: tf.SimpleDateFormat();

gaurangkudale: Hi @ramawatraksha75  Can you please explain a bit more whats the issue is?

ramawatraksha75: Sure

ramawatraksha75: As I can see we have created the method for treadSafeFormatter but didn't called the function present  in that class

gaurangkudale:
Hi @cyrille-artho 
Can you please tell me more I didnt get that

src/classes/java/lang/ThreadLocal.java
gaurangkudale: I guess it is not necessary

ramawatraksha75: Can you please help me out with that

cyrille-artho: This test is maybe not very strong in that it only checks the existence of the alternative constructor, but it is sufficient to check that the old model class of JPF did not support it and the new one does. So there is no need right now to change it.

gaurangkudale:
Hi @cyrille-artho
I was busy with academy work so that I didn't get the time for this issue now I'll continue to contribute

cyrille-artho:
Hi Guarang,
Thank you for the update. You need to add a few more `import` statements for the code to compile, as the classes `Collectors` and `List` are not in the default package. So you need to add `import java.util.List` etc., until the code compiles and passes the tests.

gaurangkudale:
Hi @cyrille-artho 
As you said I've done some changes and tests are also passed please note that waiting for your reply 
thank you

cyrille-artho:
Great, this unit test passes and is functional, so I will merge this PR and close #302.

gaurangkudale:
Excited to Do more contribution 
@cyrille-artho  What should I do next?

PR #320, opened 2022-02-09
Upgrading to gradle 7.4

gaurangkudale:
C:\Users\gaura\jpf\jpf-core>gradle clean build

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\gaura\jpf\jpf-core\build.gradle' line: 24

* What went wrong:
A problem occurred evaluating root project 'jpf-core'.
> Could not find method testCompile() for arguments [org.ow2.asm:asm:7.1] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

getting this error after trying to run `gradlew clean build` 
`testCompile` is not supported from Gradle version 7.0 alternatively I tried to change it to `taskImplimantation` but after that, it showed another exception which is mentioned in the log file below

* Get more help at https://help.gradle.org
[log.txt](https://github.com/javapathfinder/jpf-core/files/8029276/log.txt)


gaurangkudale:
Hi @cyrille-artho 
I've alternatively tried to change the `testImplementation` to `compileClasspath` but it didn't workokout

cyrille-artho:
I see; it seems there are major changes moving to Gradle version 7. There is an issue on Gradle describing this:
https://github.com/gradle/gradle/issues/15661 
The new dependency mechanism is described here: https://docs.gradle.org/current/userguide/declaring_dependencies.html
Moving to Gradle 7 therefore requires changing the way dependencies are declared.

gaurangkudale:
Hi @cyrille-artho 
I've done some changes and the tests are also passing but there are some dependencies warnings that we need to address and also getting the same results as we were getting for the current version(6.9)


[Gradle version 7.4 Logs.txt](https://github.com/javapathfinder/jpf-core/files/8037833/Gradle.version.7.4.Logs.txt)

gaurangkudale:
`Task ':compileClassesJava' uses this output of task ':copyResources' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.`

I've one doubt here in the above statement `compileClassesJava` is depends on `copyResources` but the `compileClassesJava` is not publically available to declare the dependencies (which comes under java plugin) so how to declare the dependency for that?

cyrille-artho:
This is indeed a problem. I was not able to find out so far how to make this dependency in `compileClasses` explicit. If the dependency model changes in Gradle 7, then we risk having an undefined dependency that was defined before.
However, without knowing more, it may even be possible that this dependency was not completely defined earlier.
Please keep looking for information (unfortunately, good documentation on Gradle is not easy to find) and keep us updated.

PR #321, opened 2022-02-13
Upgrading to gradle version 6.9

gaurangkudale:
I've done some changes to upgrade the Gradle version 5.4.1 to 6.9


cyrille-artho:
This looks like the Java version is not configured correctly. Perhaps you have used the exact same settings as on the main branch? In that case, Java 8 would be used, which does not work against code written for Java 10 or 11.

gaurangkudale:
> This looks like the Java version is not configured correctly. Perhaps you have used the exact same settings as on the main branch? In that case, Java 8 would be used, which does not work against code written for Java 10 or 11.

First I used java 8 but it was showing me same compilation error after that I switch to java 11 at that time it is showing me same

gaurangkudale:
For the main Branch I've used java 8 not 11

cyrille-artho:
I think a patch would also need a modified copy of `.github/workflows/simple_build.yml` as Travis no longer works for us.
The version of `.github/workflows/simple_build.yml` on the master branch uses Java 8. Maybe it can be adapted by simply changing version 8 to version 11? Could you please try that?

gaurangkudale:
Yup, I'll try my best

PR #326, opened 2022-04-09
Save properties in VM

### LINKED ISSUE #325: A Possible Solution for testSimpleReadbackOk
Currently, the test "testSimpleReadbackOk" fails in the Java 8 version of JPF. This is caused by an empty "savedProps" in sun.misc.VM. This field is set by VM.saveAndRemoveProperties(), which is only called in System.initializeSystemClass().

However, initializeSystemClass() is a private method that is called by a native library, and in my tests JPF never calls this method. A possible fix for this would be to run this method at <clinit> of System.class.

Please note that this does not fix the Java 11 version of this test, but instead the Java 8 version, which some users have reported  to be failing (#314).

ThisTestUser:
This fixes #325, as the testSimpleReadbackOk runs correctly in Java 8 after this patch.

Note that this is not the complete implementation in initializeSystemClass(), but the minimum of what is required to fix this test case.

### DIFF
--- a/src/classes/java/lang/System.java
+++ b/src/classes/java/lang/System.java
@@ -61,6 +61,7 @@ public class System {
         properties.put(kv[i], kv[i+1]);
       }
     }
+    sun.misc.VM.saveAndRemoveProperties(properties);
 
     // this is the Java 6 sun.misc.SharedSecrets backdoor mechanism which I
     // would have prefered not to learn about. It's a mess WRT Java 1.5 / 6 compatibility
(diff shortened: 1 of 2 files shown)


cyrille-artho:
Hi,
Thank you for the pull request. With this test, `testSimpleReadbackOk` succeeds, but a different test fails:

```
gov.nasa.jpf.vm.multiProcess.TypeSeparationTest > systemClassLoaderTest FAILED
    java.lang.AssertionError at TypeSeparationTest.java:102
```

ThisTestUser:
I managed to look into this, and in the process fixed an obscure bug at ClassInfo. The bug is where cached interfaces would have the old Classloader when the clone() method is called.

One thing to note is that testSimpleReadbackOk does initialize sun.misc.VM, which isn't implemented here. This is the only way to fix the bug though, which I think only happens on Windows machines.

cyrille-artho:
Fantastic, these bugs are really hard to spot, as a different test case execution order may hide such a bug. Merging now.

Issue #258, opened 2020-08-14
JPF does not call toString when preparing data for string concatenation


amgad-rady:
The following code will not print "What is this? This is an instance of X" if run in JPF. The toString method is not called properly during string concatenation.
```lang=java
class X {
  public String toString() {return "This is an instance of X";}
}

public class Concat {
  public static void main(String[] args) {
    X x = new X();
    System.out.println("What is this? " + x);
  }
}
```

Jaikishan-Saroj:
Can i Work on this issue ?? @amgad-rady

Jash-Mehta:
Can i Work on this issue?? @amgad-rady

Jash-Mehta:
Can you tell me in detail. About this issue.
@amgad-rady

cyrille-artho:
As far as I know, Amgad is not working on this issue right now.
The issue is that when concatenating objects that are not strings (and on which toString() has not been called), the VM itself automatically calls toString() in Java 9 and higher. This has to be implemented in the code that prepares the argument for the string concatenation.
To do this, the correct toString() method of each object needs to be looked up (unless that object is already a string), and it has to be called. I think there is some code elsewhere in JPF that performs a method call (at the JPF VM level) from inside JPF.
Implementing this requires a very good understanding of how JPF works, so this issue is far from ideal for getting started with JPF.

Jash-Mehta:
I think in this issue  we have to create Concat class. Otherwise it show an error like you have not create Concat class.

Jash-Mehta:
I think in this issue we have to create another class of java which is concat and call that concat class at X class(main)..
If the stuff I've done is valid according to you, I'll send a PR for it.

Jash-Mehta:
I think in this issue we have to create another class of java which is concat and call that concat class at X class(main)..
If the stuff I've done is valid according to you, I'll send a PR for it.
@cyrille-artho

cyrille-artho:
There is already code that deals with string concatenation. It is rather complex, because this happens inside a bootstrap method, which has to be correctly recognized and invoked at runtime. See `src/main/gov/nasa/jpf/vm/FunctionObjectFactory.java`, `makeConcatWithConstants`, `makeConcatWithStrings`, and the corresponding documentation of Java 11+ (`BootstrapMethodInfo` and related classes).

PR #327, opened 2022-04-09
Add failing string concat tests (Java 11)

ThisTestUser:
Add string concatenation tests involving toString(), which currrently fail in Java 11, as the object itself is passed into StringConcatFactory.makeConcatWithConstants() during the invokedynamic call. This issue was described in #258.


cyrille-artho:
Thank you for the tests. I will merge the PR but have realized that GitHub actions were not yet set up for this branch. I'll do that first, and I hope I can run this PR on the CI setup retroactively (perhaps it has to be closed and re-opened?).

PR #329, opened 2022-04-20
Add Gradle support for the java-8 branch

glockyco:
Backports the Gradle configuration from the java-10-gradle branch to the java-8[-gradle] branch.

I've also updated the .gitignore file to match the current master and removed the now-ignored IDE configuration files, as was done previously for the non-java-8 branches with commits 6b734c1aecd121c7fcaca98a99d43f34fedefdfa and c0c3ef2475f5c20378321e0dc2069744e1a63462.

The GitHub Actions configuration assumes that the target branch will be called "java-8-gradle" rather than "java-8", thereby matching the naming of the "java-10-gradle" branch.

One of the tests initially failed due to a missing method in one of the model classes (I assume the test also failed before adding Gradle support?), though that was simple to fix by adding the missing method to the model class.


cyrille-artho:
Thank you for looking into this.
Your PR shows that we have outdated branches that we should update or delete:

* Java 8 support is still on branch `master`, and branch `java-8` should probably be (primarily) mirrored from master, and be updated against `master`.
* A few enhancements of `java-8-backport` should be merged into `java-8` and `master` as well.
* `java-10` is likely obsolete, as most developments were made in `java-10-gradle`, which should be renamed to `java-11`.

The problem is that some people may still have pending tasks on these branches, whose names were accurate when they were created but are no longer ideal now (or the branch is outdated in the first place).

What do you think about the following:

* Accept your PR once all tests pass (this may require closing this PR and re-opening it so the `.github` workflow is used).
* Merge `master` and `java-8-backport`.
* We want to consider closing `java-8` and reopening it once Java 11 support works for all string concatenation cases (see #258), as Java 11 could become the default then. (Ideally, we switch once 100 % of the unit tests pass, but this may require more time.)

glockyco:
Thanks for the quick response!

> Java 8 support is still on branch `master`, and branch `java-8` should probably be (primarily) mirrored from `master`, and be updated against `master`.

Oh, so the branch that's supposed to be used when Java 8 support is required is actually `master` rather than `java-8`? And the `java-8` branch should ideally just point to the same commit as `master`, at least until `master` moves to Java 11? And the `java-8` branch is only kept alive in the first place because "people may still have pending tasks on [it]"?

Having worked primarily with `jpf-symbc` / SPF rather than just `jpf-core` / JPF, I was always under the impression that SPF, which [explicitly states that it doesn't support versions beyond Java 8](https://github.com/SymbolicPathFinder/jpf-symbc/blob/ede6f17ef950c6f71ffa025499033cdc6fc749b7/README.md?plain=1#L25), wouldn't work with anything newer than the `java-8` branch, as suggested by [some advice in the mailing list](https://groups.google.com/g/java-pathfinder/c/ufuB_bgx9ZM/m/OcxO2lGABAAJ), for example.

However, given that `master` still supports Java 8, SPF should probably also still work with `master`? Definitely going to try this, though the `jpf-symbc` repository is probably a better place to discuss this than here.

---

> What do you think about the following:
>
> - Accept your PR once all tests pass (this may require closing this PR and re-opening it so the .github workflow is used).

Assuming that my understanding above is correct, there's probably no need for this PR to exist, right? After all, there's already a branch with Java 8 AND Gradle support: `master`. If so, I'm 100% fine with the PR just being closed. No need to create more busywork for little to no benefit, especially if `java-8` in its current form is going to cease existing "soon" anyway.

> Merge `master` and `java-8-backport`

Having only worked with the `java-8` branch so far, I'm not sure I'm qualified to judge the usefulness / necessity of merging `master` and `java-8-backport`.

Having said that: if the improvements made in the Java 10/11/12 branches might also be beneficial for the Java 8 / `master` branch, backporting of these changes probably wouldn't be a bad idea. Especially if Java 8 support (albeit in a non-master branch) is something that should be offered by `jpf-core` for the foreseeable future anyway.

I might be able to squeeze in a few hours of work on that merging / backporting effort, though I don't want to promise too much here. Just let me know if you think that's a suitable task for a "JPF newbie" such as myself. If there's anything I should know before getting started on a `master`-`java-8-backport`-merge, I'd be happy to discuss that in a new issue or so.

cyrille-artho:
Thanks, I'll look into closing branch `java-8` for now. I also think it's probably better to backport to `master` rather than `java-8-backport`.

glockyco:
I've now tested `jpf-symbc` and `jpf-nhandler` with `jpf-core:master` rather than `jpf-core:java-8`. The result was that all tests of both projects still pass when using the `master` branch. I also did some additional manual testing on my own example programs and didn't observe any issues there either.

So, as far as I can tell, `master` should really be preferred over `java-8` when Java 8 support is needed.

Based on this observation, I'd suggest to:

1. Remove the `java-8` branch until `master` moves to a newer Java version.
2. Specify in the READMEs of the different branches (especially the `master` branch) which Java version(s) they support.
3. Optionally, also include a full overview of all "officially supported"/"active" branches and their supported Java versions in the READMEs. That way, people will know where to go to if they need support for a specific Java version.

glockyco:
Note that the wiki still mentions the `java-8` branch [at a few places](https://github.com/javapathfinder/jpf-wiki-sync/search?q=java-8), stating that Ant support is provided there. If this (i.e., some branch which officially supports Ant) is something that should be kept, perhaps introducing a new `java-8-ant` branch might be the way to go.

cyrille-artho:
We are about to move to Java 11 in a few weeks, and `master` already has Gradle support. So we will not update this branch anymore.

Issue #274, opened 2020-10-26
Java 11 support


cyrille-artho:
There is a branch `java-10-gradle` that we use for work to support Java 11 (that started with Java 10 support).

To reproduce the current state, try to install OpenJDK version 11 and compile the code from that branch, and run the tests with `./gradlew check`.

Currently 16 unit tests are still failing there. Fixes are very welcome. Some of these problems below are already referenced in an issue. If someone starts to work on a particular test, please create a separate issue referring to one or two of these failing tests:

```
    CountDownLatchTest. testCountDown
    ExecutorServiceTest. testShutdown
    SemaphoreTest. testResourceAcquisition
    ObjectStreamTest. testSimpleReadbackOk
    RuntimeTest. testAvailableProcessors
    StackWalkerTest. testCallerClass
    URLClassLoaderTest. testThrownException
    DateFormatTest. testConversionCycle
    DateFormatTest. testFormatWithTimeZone
    DateFormatTest. testSetAndGetTimeZone
    SimpleDateFormatTest. testFormatWithTimeZone
    RecursiveClinitTest. testNewInstance
    ProxyTest. testAnnoProxy
    ProxyTest. testBasicProxy
    InterruptTest. testPark
    SAXParserTest. testSimpleParse 
```

cyrille-artho:
### How to help with failing tests for Java-11 support

1. Make sure Java 11 is enabled by default, e.g., 
```
export JAVA_HOME="`/usr/libexec/java_home -v11.0.2`"
```
on Mac OS X, similar setting on Linux.

2. Build and run the unit tests:
```
./gradlew check
```

3. Look at the overall summary in `build/reports/tests/test/index.html` for the summary you see above in this post.

4. To look at a particular unit test failure, look at the XML report, e.g.,
```
less build/test-results/test/TEST-gov.nasa.jpf.test.java.concurrent.CountDownLatchTest.xml
```

Parts of the log relating to the failure:

```
    <failure message="java.lang.AssertionError: JPF found unexpected errors: gov.nasa.jpf.vm.NoUncaughtExceptionsProperty" type="java.lang.AssertionError">java.lang.AssertionError: JPF found unexpected errors: gov.nasa.jpf.vm.NoUncaughtExceptionsProperty
        at gov.nasa.jpf.util.test.TestJPF.fail(TestJPF.java:164)
        at gov.nasa.jpf.util.test.TestJPF.noPropertyViolation(TestJPF.java:816)
[...]

PR #350, opened 2023-05-17
Fix string concat bug of not calling correct toString() method

quadhier:
This PR should fix the following two failing string concat tests, which are parts of #274.
```
java11.StringConcatenationTest::testStringConcatenationWith_toString
java11.StringConcatenationTest::testStringConcatenationWith_toStringNested
```
I have tested locally and find that it fixes these two failing tests with no more regressions (11 failing tests left now).
## Problem Analysis
From JDK 9, the default implementation of string concat is changed to use `invokedynamic`. In OpenJDK's implementation, during the execution of `invokedynamic`, it generates bytecode dynamically to convert the args of string concat to the right data type (for example, call `toString()` of the reference type) and concatenate them.

In the current implementation of JPF, during using `invokedynamic` to do string concat, it doesn't call the `toString()` method of reference types in the string concat args. It could cause the result string to be wrong in a case like `"Info: " + Person` where `Person` has a user-defined `toString()` method.

## Challenges
1. Handwriting bytecode is hard, but importing libraries like `asm` is too heavy.
2. JPF is a model checking JVM saving program states; generating too much or too complex code could bring unnecessary runtime burdens.

## Basic Idea of This Patch
In the case of string concat, `invokedynamic` works like a function that receives variable-length args and returns a reference to a concatenated String. Similarly, this patch implements the following:
1. During the execution of `invokedynamic` itself, the args of string concat are collected and put into a reference array, a new stack frame is created to call the helper function `java.lang.String.generateStringByConcatenatingArgs()`.
2. Then, the execution continues from the helper function. It uses `StringBuilder` to concatenate the args of `invokedynamic` and return a reference to the concatenated String.

## Limitations
It cannot handle cases where string concat args have special string literals containing `'\u0001'` or `'\u0002'`. In these cases, BSM could have more than one static argument (please refer to [this SO answer](https://stackoverflow.com/a/65183722)). Even JPF's class file parser has not yet supported this case. So classfiles containing these cases can make JPF fail at the class loading phase. Fortunately, these are rare cases. I'd like to add support to them later.

### DIFF
--- a/src/classes/modules/java.base/java/lang/String.java
+++ b/src/classes/modules/java.base/java/lang/String.java
@@ -468,4 +468,24 @@ implements java.io.Serializable, Comparable<String>, CharSequence {
   }
 
   native public static void checkBoundsOffCount(int offset, int count, int length);
+
+
+  // This private method is NOT part of the String API.
+  // It is meant to be used ONLY in FunctionObjectFactory
+  // for invokedynamic's String concatenation.
+  //
+  // Any further modifications of this function MUST NOT use any
+  // string concatenation operations in it, otherwise it can
+  // trap into infinite recursion.
+  private static String generateStringByConcatenatingArgs(Object[] args) {
+    // We use StringBuilder just like what OpenJDK does before JDK 9.
+    // More importantly, unlike StringBuffer, StringBuilder
+    // doesn't involves any synchronization operations, which prevent JPF
+    // from generating unnecessary state transitions.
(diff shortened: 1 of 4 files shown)


cyrille-artho:
Looks good to me! Pavel?

quadhier:
Thanks for your review, Cyrille!

pparizek:
I like the PR too.

quadhier:
Thanks a lot, Pavel!

PR #353, opened 2023-05-20
Fix a not-throwing-exception bug in Class.newInstance

quadhier:
This PR should fix a bug that, in some cases, execution should throw an exception, but it doesn't on JPF. And a test case for this bug is added. This patch may also apply to `master` branch (on OpenJDK 8).

I have tested locally on `java-10-gradle` branch; this fix brings no regressions (still, 11 failing tests left). 

## Reproduction
According to Java API doc for [java.lang.Class.newInstance()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#newInstance()), the following code snippet should throw `InstantiationException`. (It falls into the category of "the class has no nullary constructor"). But it doesn't throw any exception on JPF.
```
public class Main {

  static class C {
    int i;

    public C(int i) {
      this.i = i;
    }
  }

  public static void main (String[] args) throws Exception {
    C.class.newInstance();
  }
}
```

## Problem Analysis
JPF doesn't check this case and does the wrong dispatch in the `java.lang.Class.newInstance` code.

https://github.com/javapathfinder/jpf-core/blob/3408119d115e539956a3d920e22e856e05bb9d23/src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java#L222-L229

## Basic Idea of This Fix
Just handle this case in the implementation of `java.lang.Class.newInstance`.

### DIFF
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java
@@ -225,7 +225,15 @@ public class JPF_java_lang_Class extends NativePeer {
     DirectCallStackFrame frame = ti.getReturnedDirectCall();
     
     ClassInfo ci = env.getReferredClassInfo(robj);   // what are we
-    MethodInfo miCtor = ci.getMethod("<init>()V", true); // note there always is one since something needs to call Object()
+    MethodInfo miCtor = ci.getMethod("<init>()V", false);
+
+    // According to Java SE 11 API Specification,
+    // If "the class has no nullary constructor", this method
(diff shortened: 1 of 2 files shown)


cyrille-artho:
Looks good, thank you for tracking this down!
Can you please make a PR against the `master` branch? That will allow the CI build to run all the tests. We can then later merge all the changes from `master` into `java-10-gradle`. (Actually, it would be good to do this from time to time to ensure we don't miss anything from there.)

quadhier:
Done. But that patch may not be directly applied to `java-10-gradle` branch because of some file differences.

PR #354, opened 2023-05-21
Fix a not-throwing-exception bug in Class.newInstance, a backport of PR #353

quadhier:
This is a backport of #353 on `master` branch. This patch is a little bit different from that one because of some file differences.


cyrille-artho (approved):
Many thanks!

PR #365, opened 2023-06-30
Add object stream support for JPF on Java 11

quadhier:
This patch should fix the issue that appears when `ObjectInputStream` reads an object thus adding object stream API support for JPF on Java 11.

It should fix the following failing test mentioned in #274 and bring no more regressions.
```
gov.nasa.jpf.test.java.io.ObjectStreamTest::testSimpleReadbackOk
```

After this patch, there might be still 5 unit test failures left in JPF on Java 11.

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/JVMClassFileContainer.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassFileContainer.java
@@ -125,7 +125,7 @@ public abstract class JVMClassFileContainer extends ClassFileContainer {
       // Here we try to load the class on the underlying JVM (OpenJDK) to
       // query its module name. But for gov.nasa.jpf.SerializationConstructor,
       // OpenJDK will see it inheriting a final class (j.l.r.Constructor) and throw
-      // java.lang.VerifyError. Since we know it doesn't have module name, we can
+      // java.lang.VerifyError. Since we know it doesn't have a module name, we can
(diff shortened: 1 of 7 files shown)


src/main/gov/nasa/jpf/jvm/JVMClassFileContainer.java
```diff
@@ -117,6 +117,20 @@ static String getModuleName(String typeName) {
...
     try {
+      // In support of jdk.internal.reflect.ReflectionFactory,
+      // we use a model class java.lang.reflect.Constructor (which is non-final
+      // in our implementation but is final in OpenJDK's implementation)
+      // and define gov.nasa.jpf.SerializationConstructor as a subclass of it.
+      //
+      // Here we try to load the class on the underlying JVM (OpenJDK) to
+      // query its module name. But for gov.nasa.jpf.SerializationConstructor,
+      // OpenJDK will see it inheriting a final class (j.l.r.Constructor) and throw
+      // java.lang.VerifyError. Since we know it doesn't have module name, we can
```
cyrille-artho, line 128: minor: "it doesn't have _a_ module name"

cyrille-artho (changes requested):
Looks good! Found a minor typo but otherwise looks good to me.

cyrille-artho:
Thanks a lot!

PR #367, opened 2023-07-10
Bytecode for REF_newInvokeSpecial: load new object as first argument

zhangwen0411:
A constructor takes the new object as its _first_ argument. This patch modifies bytecode generation for `REF_NEW_INVOKESPECIAL` to load the new object _before_ any other arguments to a constructor.

Submitting pull request to the `java-10-gradle` branch, where the previous fix (#359) is located.


cyrille-artho:
Hi,
Thank you very much for spotting this and creating a PR with a new test case.
Your PR looks good; the CI pipeline fails because the build was made against an older version. I'll merge it and assume it will work against the new version (other than the SAX Parser test).

PR #372, opened 2023-07-18
Fixed Class.forName for primitive array types

zhangwen0411:
For example, in the current implementation `Class.forName("[B")` fails with:

> java.lang.ClassNotFoundException: class not found: B

which should be fixed by this patch. The type resolution is based on that of `INSTANCEOF.execute` ([here](https://github.com/javapathfinder/jpf-core/blob/1f99a23a095e89b22b71539dc238ba5d73231a6d/src/main/gov/nasa/jpf/jvm/bytecode/INSTANCEOF.java#L40) and [here](https://github.com/javapathfinder/jpf-core/blob/1f99a23a095e89b22b71539dc238ba5d73231a6d/src/main/gov/nasa/jpf/jvm/bytecode/INSTANCEOF.java#L45-L60)).

(Incidentally, `Class.forName("[B")` does work in unit tests because [there exists](https://github.com/javapathfinder/jpf-core/blob/1f99a23a095e89b22b71539dc238ba5d73231a6d/src/tests/TypeNameTest.java#L31) a class named `B` in the unnamed package.)

### DIFF
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java
@@ -188,27 +188,22 @@ public class JPF_java_lang_Class extends NativePeer {
     // class of the method that includes the invocation of Class.forName() 
     ClassInfo cls = mi.getClassInfo();
 
-    String name;
-    // for array type, the component terminal must be resolved
-    if(clsName.charAt(0)=='[') {
-      name = Types.getComponentTerminal(clsName);
-    } else{
-      name = clsName;
-    }
-
-    // make the classloader of the class including the invocation of 
-    // Class.forName() resolve the class with the given name
-    try {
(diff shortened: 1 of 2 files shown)


cyrille-artho:
Great, this fixes the bug indeed! Is there an issue relating to it?

cyrille-artho (approved):

zhangwen0411:
I'm not aware of any issue relating to this!

cyrille-artho:
@pparizek : This patch looks good to me, can you please take a look also?
@zhangwen0411 : Thank you for finding and fixing this. Please keep your branch in your repo (once we merge your PR) as this PR can probably also be made against branch `java-10-gradle`, which will soon fully support Java 11 for JPF.

pparizek:
I agree with this fix.

cyrille-artho:
Thank you, and please make the same PR against `java-10-gradle` (if the patch makes sense for that branch also).

PR #376, opened 2023-07-20
gov.nasa.jpf.vm.Allocation.equals: added missing equality check for the count field

zhangwen0411:
Without this check, two `Allocation` objects with the same context, same hash, but different counts will be erroneously treated as equal.


cyrille-artho (approved):

cyrille-artho:
@zhangwen0411 : Thank you again. If Pavel approves this PR, please also make it against branch `java-10-gradle` if it works there.
@pparizek : Looks good to me. Can you please also check?

pparizek (approved):

PR #377, opened 2023-07-20
gov.nasa.jpf.vm.Allocation.equals: added missing equality check for the count field

zhangwen0411:
Same as #376, but against the `java-10-gradle` branch.


cyrille-artho:
Many thanks!

PR #381, opened 2023-08-02
Fix bugs in some calculations related to field sizes of function object created for `invokedynamic` (#378)

### LINKED ISSUE #378: Lambda expression fails to capture a double type local variable in JPF on Java 11
## Problem
This program will cause JPF on Java 11 (`java-10-gradle` branch) to crash:
```java
interface I {
  void foo();
}

public class Test {
  public static void main (String[] args) {
    double d = 0.0;
    I i = () -> {
      System.out.println(d);
    }
  }
}
```

## Bug Diagnosis
Field created here wrongly use type name. It should use type signature.

https://github.com/javapathfinder/jpf-core/blob/6f6500c6b984b594c0244d373e51400822ac04b0/src/main/gov/nasa/jpf/vm/ClassInfo.java#L731

## A Tentative Fix

```diff
@@ -728,7 +728,8 @@ public class ClassInfo extends InfoObject implements Iterable<MethodInfo>, Gener

    int i = 0;
    for(String type: fieldTypesName) {
-     FieldInfo fi = FieldInfo.create("arg" + i++, type, 0);
+     String typeSig = Types.getTypeSignature(type, false);
+     FieldInfo fi = FieldInfo.create("arg" + i++, typeSig, 0);
      fi.linkToClass(this, idx, off);
      iFields[idx++] = fi;
      off += fi.getStorageSize();
```

quadhier:
This patch should fix #378 and some other bugs in lambda expression's capturing local primitive types of different storage sizes (on branch `java-10-gradle`). A unit test is also added.

I have run `./gradlew clean check` locally and find no regressions.

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
@@ -785,7 +785,15 @@ public class JVMClassInfo extends ClassInfo {
     MethodInfo fiMethod = funcInterface.getInterfaceAbstractMethod();
     int modifiers = fiMethod.getModifiers() & (~Modifier.ABSTRACT);
     int nLocals = fiMethod.getArgumentsSize();
-    int nOperands = this.nInstanceFields + nLocals;
+    // A field may occupy 1 or 2 operand stack slots.
+    // Since all these fields need to be pushed into operand stack,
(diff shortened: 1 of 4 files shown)


cyrille-artho:
Great, thank you very much!

PR #384, opened 2023-08-03
Class.getField should return only public fields

zhangwen0411:
See, e.g., [Java 8 documentation on `Class.getField`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getField-java.lang.String-).

(Same for `Class.getMethod`—which is already [correctly implemented](https://github.com/javapathfinder/jpf-core/blob/4735817ea71fceeea68aa7eb4bd3568b2b7afa25/src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java#L350) in its JPF native peer.)


src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java
```diff
@@ -643,12 +643,12 @@ int getField (MJIEnv env, int clsRef, int nameRef, boolean isRecursiveLookup) {
   
   @MJI
   public int getDeclaredField__Ljava_lang_String_2__Ljava_lang_reflect_Field_2 (MJIEnv env, int clsRef, int nameRef) {
-    return getField(env,clsRef,nameRef, false);
+    return getField(env,clsRef,nameRef, false, false);
```
cyrille-artho, line 646: Could you please add a comment explaining why parameter `publicOnly` is set to `false` here?

cyrille-artho (changes requested):
Thank you very much for spotting this and for the thorough fix with the unit tests.
When looking at the documentation of `getField` and `getDeclaredField`, they seem to behave analogously w.r.t. access permissions. Both should only permit access to visible fields. However, you use different settings in your code. Could you please explain why?
Other than that, your patch looks great, and I'm happy to access it (both on `master` and `java-10-gradle`.

zhangwen0411:
Thank you for the review! Regarding the difference in access checks between `Class.getField` and `Class.getDeclaredField`, the Java 8 documentation says the following.

[For `getField`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getField-java.lang.String-):

> Returns a Field object that reflects the specified **public** member field of the class or interface represented by this Class object.

(emphasis mine)

[For `getDeclaredField`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredField-java.lang.String-):
> Returns a Field object that reflects the specified declared field of the class or interface represented by this Class object.

(There is no restriction of the Field object reflecting a _public_ field.)

I also tried this out on the OpenJDK 8 JVM with this example:

```java
import java.lang.reflect.Field;

public class GetDeclaredFieldExample {
    public static void main(String[] args) throws NoSuchFieldException {
        Field f = AnotherClass.class.getDeclaredField("privateField");
        System.out.println("getDeclaredField succeeded: " + f.getName());

        try {
            AnotherClass.class.getField("privateField");
        } catch (NoSuchFieldException e) {
            System.out.println("getField failed: " + e);
        }
    }
}

class AnotherClass {
    private int privateField;
}
```

It prints out:
```
getDeclaredField succeeded: privateField
getField failed: java.lang.NoSuchFieldException: privateField
```

I added comments to the `getField` and `getDeclaredField` peers regarding this difference. Once this PR gets approved, I will submit another to `java-10-gradle` branch!

cyrille-artho (approved):
Great, thanks for the clarifying comments!

PR #385, opened 2023-08-04
Class.getField should return only public fields

zhangwen0411:
* Class.getField should return only public fields

* Added comments on the behavior of `Class.getField` and `Class.getDeclaredField`

(Same as #384)


cyrille-artho:
Many thanks!

PR #386, opened 2023-08-05
More complete implementation of Field.isAccessible

zhangwen0411:
This PR:

* Fixes issue: `final` fields cannot be read via reflection because `isAccessible` returns `false` on final fields regardless of whether the access is a read or a write: https://github.com/javapathfinder/jpf-core/blob/f0e39060944bff59ef5a970ad155bd82dcbb265f/src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_Field.java#L102-L104
* Contains a more complete implementation of reflection field-access checking.

I _believe_ this implementation is complete with respect to OpenJDK 8, except for two behaviors:

1. OpenJDK 8 seems to disallow writing to a static final field using reflection, even if overridden with `setAccessible(true)` (https://github.com/openjdk/jdk8u/blob/0bec498fc945a49f1134e02eb477e0af15db3b5b/jdk/src/share/classes/sun/reflect/UnsafeFieldAccessorFactory.java#L38). This implementation allows such accesses because they occur in tests: e.g., https://github.com/javapathfinder/jpf-core/blob/f0e39060944bff59ef5a970ad155bd82dcbb265f/src/tests/gov/nasa/jpf/test/mc/basic/FinalBreakTest.java#L162-L167

2. OpenJDK 8 seems to disallow using reflection to access a protected static field of a parent class from a different package, like this:
```java
@Test
public void getParentProtStatic() throws NoSuchFieldException, IllegalAccessException {
    Field f = PublicParent.class.getDeclaredField("parentProtectedField");
    Object value = f.get(new PublicParent()); // <- throws IllegalAccessException
}
```
But this behavior is allowed by OpenJDK 11, and allowing it seems to make sense, so this implementation allows it.

### DIFF
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_Field.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_Field.java
@@ -90,51 +90,84 @@ public class JPF_java_lang_reflect_Field extends NativePeer {
     StackFrame frame = ti.getTopFrame(); // this is the Field.get/setX(), so we still have to get down
     return frame.getPrevious();
   }
-  
-  protected boolean isAccessible (MJIEnv env, FieldInfo fi, int fieldRef, int ownerRef){
-    
-    // note that setAccessible() even overrides final
-    ElementInfo fei = env.getElementInfo(fieldRef);
-    if (fei.getBooleanField("isAccessible")){
+
+  // Roughly follows the logic implemented in `sun.reflect.Reflection.verifyMemberAccess` of OpenJDK 8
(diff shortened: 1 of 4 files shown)


cyrille-artho:
Thanks, this looks good! @pparizek, can you please also review this patch? It is quite long so I may have missed something.

pparizek (approved):

pparizek:
I like the patch. You can merge it.

cyrille-artho:
Thanks, merged. @zhangwen0411 , please issue this patch to `java-10-gradle` as well.

Issue #383, opened 2023-08-03
Class defined on custom ClassLoader is not recognized


zhangwen0411:
# What's the problem
Consider this test case: https://github.com/javapathfinder/jpf-core/compare/master...zhangwen0411:jpf-core-public:classloader_hierarchy_test#diff-b5a365fbf63a28037992584d9a8d8de725e7ba82b5246bd60400695c1e885bfdR213; an excerpt is here:
```java
  @Test
  public void testDefineClassInstanceOf() throws NoSuchMethodException, InvocationTargetException, InstantiationException, IllegalAccessException {
    if (verifyNoPropertyViolation()) {
      TestClassLoader testClassLoader = new TestClassLoader();
      testClassLoader.defineClassWithDefaultConstructor("example/MyClass1");
      Class<?> cls2 = testClassLoader.defineClassWithDefaultConstructor("example/MyClass2");
      Object obj = cls2.getConstructor().newInstance();
      boolean objIsInstanceOfObject = obj instanceof Object;
      assertTrue(objIsInstanceOfObject);
    }
  }
```

The test case fails on this line (before the actual `assertTrue`):
```java
boolean objIsInstanceOfObject = obj instanceof Object;
```
with:
```
====================================================== error 1
gov.nasa.jpf.vm.NoUncaughtExceptionsProperty
java.lang.ClassNotFoundException: class not found: example.MyClass2
    at gov.nasa.jpf.test.java.lang.ClassLoaderTest.testDefineClassInstanceOf(ClassLoaderTest.java:221)
    at java.lang.reflect.Method.invoke(gov.nasa.jpf.vm.JPF_java_lang_reflect_Method)
    at gov.nasa.jpf.util.test.TestJPF.runTestMethod(TestJPF.java:648)
```

The JPF stack trace looks like:
```
gov.nasa.jpf.vm.ClassInfoException: class not found: example.MyClass2
    at gov.nasa.jpf.vm.ClassLoaderInfo.getResolvedClassInfo(ClassLoaderInfo.java:363)
    at gov.nasa.jpf.vm.SystemClassLoaderInfo.getResolvedClassInfo(SystemClassLoaderInfo.java:147)
    at gov.nasa.jpf.vm.ClassLoaderInfo.getCurrentResolvedClassInfo(ClassLoaderInfo.java:149)
    at gov.nasa.jpf.vm.Types.instanceOf(Types.java:1091)
[...]

PR #388, opened 2023-08-10
Modify Types.instanceOf to take a ClassInfo as LHS

zhangwen0411:
This method's declaration currently reads `instanceOf(String, String)`, taking both the LHS and RHS types as type signature strings.  As a result, to use a `ClassInfo` object `ci` as the LHS, a caller must convert it into a string via `ci.getType()` (see `ElementInfo.instanceOf`), before the string eventually gets turned back into a `ClassInfo` object (if it's a reference type) in the method body.

This patch changes the method declaration to read `instanceOf(ClassInfo, String)`.  This change not only saves the back-and-forth, but also addresses the case where the LHS class has been loaded by a non-default class loader, in which case the original implementation fails to resolve its type string back into a `ClassInfo`.  Fixes https://github.com/javapathfinder/jpf-core/issues/383.

This patch also simplifies the implementation of `instanceOf`, ensures it recognizes arrays as subtypes of Cloneable and Serializable (JLS 4.10.3), changes the names of a few variables to clarify they're expected to contain type signatures, and adds unit tests (to `ClassTest.java`, where the existing `testInstanceOf` unit test resides).

One downside is that if LHS is a nested array, the new implementation will resolve a `ClassInfo` object for every "layer" of the array (via `ClassInfo.getComponentClassInfo`), while the original implementation "peels" the array via (cheaper) calls to `String.substring`.  I'm guessing this performance degradation will not be noticeable in common scenarios.

### DIFF
--- a/src/main/gov/nasa/jpf/vm/Types.java
+++ b/src/main/gov/nasa/jpf/vm/Types.java
@@ -1070,30 +1070,18 @@ public class Types {
     return (int) (l >> 32);
   }
 
-  public static boolean instanceOf (String type, String ofType) {
-    int bType = getBuiltinTypeFromSignature(type);
-
-    if ((bType == T_ARRAY) && ofType.equals("Ljava/lang/Object;")) {
-      return true;
+  public static boolean instanceOf (ClassInfo ci, String ofTypeSignature) {
+    int bOfType = getBuiltinTypeFromSignature(ofTypeSignature);
+    if (ci.isArray() && bOfType == T_ARRAY) {
+      return instanceOf(ci.getComponentClassInfo(), getArrayElementType(ofTypeSignature));
     }
 
-    int bOfType = getBuiltinTypeFromSignature(ofType);
-
-    if (bType != bOfType) {
-      return false;
+    if (ci.isPrimitive()) {
+      int bType = getBuiltinTypeFromSignature(ci.getSignature());
+      return bType == bOfType;
     }
 
-    switch (bType) {
-    case T_ARRAY:
-      return instanceOf(type.substring(1), ofType.substring(1));
-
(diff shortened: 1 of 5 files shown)


src/main/gov/nasa/jpf/vm/Types.java
```diff
@@ -1070,30 +1070,18 @@ public static int hiLong (long l) {
...
     }
 
-    int bOfType = getBuiltinTypeFromSignature(ofType);
-
-    if (bType != bOfType) {
-      return false;
+    if (ci.isPrimitive()) {
+      int bType = getBuiltinTypeFromSignature(ci.getSignature());
+      return bType == bOfType;
     }
```
zhangwen0411, line 1082: This special case for `ci.isPrimitive()` shouldn't be necessary, but I was confused as to what `ClassInfo.isInstanceOf` expects as argument:
https://github.com/javapathfinder/jpf-core/blob/8ae607ae65e3f5e92a04963157ee5315202ecefc/src/main/gov/nasa/jpf/vm/ClassInfo.java#L1620-L1629

When the `ClassInfo` is not primitive (the `else` branch), this method seems to expect `cname` to be a class name or a type name (as exemplified in the usage `ci.isInstanceOf(getTypeName(ofType))` in the old code). But for a primitive type (the `if` branch), it expects `cname` to be a type signature (e.g., `I` rather than `int`). So either way I had to special case it.

cyrille-artho (approved):
Again, LGMT, @pparizek : Please check also.

pparizek:
I like the patch too.

pparizek (approved):

cyrille-artho:
Please also apply this PR to branch `java-10-gradle` if it makes sense there.

PR #391, opened 2023-08-11
Add serializable lambda expression support (#234)

### LINKED ISSUE #234: JPF fails to correctly type the return values of certain method calls
The method Comparator.comparingInt returns a lambda expression cast to java.util.Comparator<T> and the interface java.io.Serializable. JPF fails to emulate this properly as its result of the comparingInt call does not implement the interface Serializable. Similar problems arise in other methods in the Comparator class.

quadhier:
This patch adds support for serializable lambda expression and also fixes #234.

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
@@ -150,7 +150,8 @@ public class JVMClassInfo extends ClassInfo {
         // The check order cannot be reversed since other BSM may not
         // have forth argument in `cpArgs`
         boolean isSerializable = false;
-        if (mth.equals("altMetafactory")) {
+        if (cls.replace('/', '.').equals("java.lang.invoke.LambdaMetafactory")
+            && mth.equals("altMetafactory")) {
           int flags = cf.intAt(cpArgs[3]);
           int FLAG_SERIALIZABLE = 1 << 0;
           if ((flags & FLAG_SERIALIZABLE) != 0) {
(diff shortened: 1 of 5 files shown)


src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
```diff
@@ -140,6 +140,24 @@ public void setBootstrapMethod (ClassFile cf, Object tag, int idx, int refKind,
...
+        // with OpenJDK 11' implementation of
+        // java.lang.invoke.LambdaMetafactory::altMetafactory()
+        //
+        // Check if it is serializable lambda expression. It is if:
+        //   1. bootstrap method is "altMetafactory"
+        //   2. 4-th cp arg value has FLAG_SERIALIZABLE (1 << 0) bit set
+        // The check order cannot be reversed since other BSM may not
+        // have forth argument in `cpArgs`
+        boolean isSerializable = false;
+        if (mth.equals("altMetafactory")) {
```
cyrille-artho, line 153: Shouldn't the class/interface be checked here, too?

cyrille-artho (changes requested):
Looks good; I think also checking the class/interface against the expected method is perhaps needed?

src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
```diff
@@ -150,7 +150,8 @@ public void setBootstrapMethod (ClassFile cf, Object tag, int idx, int refKind,
         // The check order cannot be reversed since other BSM may not
         // have forth argument in `cpArgs`
         boolean isSerializable = false;
-        if (mth.equals("altMetafactory")) {
+        if (cls.replace('/', '.').equals("java.lang.invoke.LambdaMetafactory")
```
cyrille-artho, line 153: if (cls.equals("java/lang/...") ...
// (I think your version is perhaps more readable, but for people who know bytecode, / as package name separators is nothing new, so it's better to avoid a search/replace when it's not needed

cyrille-artho (changes requested):

cyrille-artho (approved):
Looks good now; waiting for @pparizek to check as well.

pparizek:
I have no objections to this pull request.

PR #397, opened 2023-08-12
Apply patch to add System.in.available()

varad64:
The original patch can be found [here](https://github.com/javapathfinder/jpf-core/commit/55f4f09d392d1f721119ec3382caeffcefc9a895).

Adds one test : `testSystemIn`

All 934 tests pass.

Thanks!

### DIFF
--- a/src/classes/modules/java.base/java/lang/System.java
+++ b/src/classes/modules/java.base/java/lang/System.java
@@ -34,6 +34,7 @@ public class System {
 
   static Properties properties;
 
+  // These are stubs that provide an InputStream instance that has no data available.
   public static InputStream in = new InputStream() {
     public int available() { return 0; }
     public int read() { return -1; }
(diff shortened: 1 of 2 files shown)


src/classes/modules/java.base/java/lang/System.java
```diff
@@ -33,7 +34,13 @@ public class System {
 
   static Properties properties;
 
-  public static InputStream in; // new BufferedInputStream(...);  // <2do> not yet
+  public static InputStream in = new InputStream() {
+    public int available() { return 0; }
```
cyrille-artho, line 38: Please add a comment here:
// These are stubs that provide an InputStream instance that has no data available.

cyrille-artho (changes requested):

src/classes/modules/java.base/java/lang/System.java
varad64: Code now updated with this comment, @cyrille-artho.
Thanks!

cyrille-artho (approved):

PR #402, opened 2023-08-12
Apply patch for #285 and #289 to implement pollution detection and to add the gradle task

varad64:
Adds pollution detection as part of #285 :
- To review: Classes `PolDetExamples` and `PolDetMain` within `tests/PolDet` use the **GPL v3** license. 

Adds the gradle task testPolDet to the CI build via `simple_build.yml`

Thanks!


cyrille-artho (approved):
We accepted this patch back in the day because they do at least crash testing, but looking at other tests that redirect the output and check it at least partially, these tests are quite weak.

PR #408, opened 2023-08-12
Apply patch for #296, to add maven-publish plugin support

varad64:
'maven-publish' plugin is used to support publishing to the local maven repository.

No new tests added.

Thanks!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -301,6 +301,7 @@ publishing {
             url = "https://plugins.gradle.org/m2/"
         }
     }
+}
     
 def PolDetJPFClasspath = "${buildDir}/tests:" + configurations.testRuntimeClasspath.findAll { it.name.endsWith('jar') && (it.name.contains('junit') || it.name.contains('hamcrest')) }.join(":")
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 plugins {
     id 'java'
+    id 'maven-publish'
 }
 
 java {
@@ -272,6 +273,31 @@ test {
                                  "${result.skippedTestCount} skipped"]
 
             println "Summary: " + summaryFields.join(", ")
+        }
+    }
+}
+
+publishing {
+    publications {
+        jpfCore(MavenPublication) {
+            groupId = 'gov.nasa'
+            artifactId = 'jpf-core'
+            artifact createJpfJar
+        }
+        jpfAnnotation(MavenPublication) {
+            groupId = 'gov.nasa'
+            artifactId = 'jpf-annotations'
+            artifact createAnnotationsJar
+        }
+        jpfClasses(MavenPublication) {
+            groupId = 'gov.nasa'
+            artifactId = 'jpf-classes'
+            artifact createJpfClassesJar
+        }
+    }
+    repositories {
+        maven {
+            url = "https://plugins.gradle.org/m2/"
         }
     }
 }


cyrille-artho (approved):

cyrille-artho:
An earlier patch for "PolDet" already modified `build.gradle`. This leads to merge conflicts with this PR. Can you please resolve it? Simply accept both changes so PolDet remains being tested.
Also, is there a way to check if `maven-publish` produces a valid artifact?

cyrille-artho (changes requested):
See my comments; there is a minor merge request, and there is no check whether something (valid) was produced. Perhaps we can add this as part of the CI actions?

varad64:
> An earlier patch for "PolDet" already modified `build.gradle`. This leads to merge conflicts with this PR. Can you please resolve it? Simply accept both changes so PolDet remains being tested. Also, is there a way to check if `maven-publish` produces a valid artifact?

Hello Cyrille, the conflict has now been resolved. 
The gradle command `./gradlew publish` can be used to execute the publishing tasks.
Although the publishing tasks have been setup correct and comply with the gradle version that the project is running, when the publish command is executed, I am presented with the following error:

maven-publish error

I will compare the functionality on the master branch and perhaps if there is a value addition to the task, we should add it to the CI build.

cyrille-artho:
Hi,
I can see that `build/jpf.jar` is recreated with that command, and three xml manifests are written to `build/publications` (in three separate subdirectories). I don't know if this is the desired outcome.

If it is, a simple shell script (you can create a new script .github/workflows/checkMaven.sh with permissiono 0755) can check this:

```
#!/bin/sh
[[ -s build/jpf.jar && -s build/publications/jpfAnnotation/pom-default.xml && -s build/publications/jpfClasses/pom-default.xml && -s build/publications/jpfCore/pom-default.xml ]]
```

The script fails if one of these four files is missing.

cyrille-artho:
For some reason, I could not see the screenshot with the problem that you get when you try to publish to the actual (non-local) maven repository. Perhaps something has be set up first. We can deal with that later as long as it's documented somewhere as an open issue.
It's still good to have the build target in this branch, so half of the problem is solved.

varad64:
> For some reason, I could not see the screenshot with the problem that you get when you try to publish to the actual (non-local) maven repository. Perhaps something has be set up first. We can deal with that later as long as it's documented somewhere as an open task in issue #290. It's still good to have the build target in this branch, so half of the problem is solved.

Yes Cyrille, I edited my comment as the error was not relevant.
At first, I was trying to execute the `./gradlew publish` command which executes all publication tasks and publishes to maven repositories, but the functionality exists to publish to maven local repository, which can be done by means of `./gradlew publishToMavenLocal` command. I understood the difference after getting some more context about the maven-plugin

varad64:
> Hi, I can see that `build/jpf.jar` is recreated with that command, and three xml manifests are written to `build/publications` (in three separate subdirectories). I don't know if this is the desired outcome.
> 
> If it is, a simple shell script (you can create a new script .github/workflows/checkMaven.sh with permission 0755) can check this:
> 
> ```
> #!/bin/sh
> [[ -s build/jpf.jar && -s build/publications/jpfAnnotation/pom-default.xml && -s build/publications/jpfClasses/pom-default.xml && -s build/publications/jpfCore/pom-default.xml ]]
> ```
> 
> The script fails if one of these four files is missing. Of course, the content could still be wrong, but at least we can check that _something_ is produced.

Sounds good @cyrille-artho. I will try with a local maven installation to understand what happens in that case since the task is supposed to publish to the local maven repository. The check looks good now but I will get more context and see if this is all we need.
Thanks!

varad64:
Hello @cyrille-artho, I have better context about the publication tasks now and I don't think there is an explicit need to validate the output of these tasks as gradle performs validation when executing these tasks. 

I tried publishing to maven's local repository and to a directory on my local file system and both tasks successfully generate the artefacts for the 3 defined publications. The output includes the build artefacts, POM files and maven-metadata files for each of the following publications:
1. `jpf-core `
2. `jpf-classes`
3. `jpf-annotations`

And so, the addition of the publishing command to the CI build script depends on if most users would want to make use of these publication tasks. If not, the users can execute the gradle publishing commands and documentation can be added to let the users know that the publishing feature is supported.

Also, `build/jpf.jar` being recreated as a prerequisite when executing either of the publishing tasks: `/.gradlew publish` or `./gradlew publishToMavenLocal` . The screenshot below shows the series of gradle tasks that are executed when `./gradlew publishToMavenLocal` and it can be seen that `createJpfJar` is one of these tasks:
publish tasks chain

Thanks!

cyrille-artho (approved):

cyrille-artho:
OK, in that case, I assume that the local packaging task fails if there is a problem. So we can merge this PR.

varad64:
Yes Cyrille, thank you!

PR #177, opened 2019-03-13
Implements a jpf example ConcurrentCount

manish3499:
The jpf example CocnurrentCount can be used to compare the performance of different jpf-core implementations/search techniques. The state-space's size can be greatly varied by changing the variable "COUNT" which further helps in making comparisons.

PR #412, opened 2023-08-17
Apply patch for #177 to add JPF example ConcurrentCount

varad64:
No new tests added.

### DIFF
--- a/src/examples/ConcurrentCount.java
+++ b/src/examples/ConcurrentCount.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2014, United States Government, as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All rights reserved.
+ *
+ * The Java Pathfinder core (jpf-core) platform is licensed under the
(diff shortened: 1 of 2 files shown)


cyrille-artho:
Actually, when reviewing this again, I can see that the program is not thread-safe, but the assertion is written incorrectly. The current setting does not terminate in a reasonable time using the standard search, and the program is poorly documented. Most documentation was given in PR #177 .
So I am not convinced that this example should be included in newer versions of JPF as of now.

varad64:
Thanks for the feedback, @cyrille-artho.
In that case, I will close this PR, as it does not offer any value in the current state.
Will reopen if the above stated concerns are resolved.

Thanks!

cyrille-artho:
Great, thanks.--Regards,Cyrille ArthoOn 17 Aug 2023, at 19:41, Varadraj Galgali ***@***.***> wrote:﻿
Thanks for the feedback, @cyrille-artho.
In that case, I will close this PR, as it does not offer any value in the current state.
Will reopen if the above stated concerns are resolved.
Thanks!

PR #208, opened 2019-07-17
Fixed missing ClassNotFoundException for invalid array types due to incomplete checks in the method Types.getClassNameFromTypeName() (#204)

rtrimana:
1) Added a fix for the method `Types.getClassNameFromTypeName()`.
2) Added a unit test to test that this fix can catch the invalid array type with a `java.lang.ClassNotFoundException`.

Please see this [issue ](https://github.com/javapathfinder/jpf-core/issues/204) for further explanation about the problem.

PR #415, opened 2023-08-22
Adds recursive lookup for ClassLoader.defineClass() and unit tests

varad64:
Hello @cyrille-artho and @pparizek,

This PR combines a set of related patches which add functionality and fix bugs. The patches involved here are #203, #208, #209 and #388, which means #410 can now be skipped as it is same as #388 minus the unit test in `ClassLoaderTest`.

The patch adds a total of 6 unit tests. All 1002 tests now pass.

Some refactors were necessary in the `build.gradle` file to comply the added dependency for **asm** and the task with the latest gradle version.

Thanks!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,17 @@ repositories {
 }
 
 dependencies {
+    testImplementation 'org.ow2.asm:asm:9.5'
     testImplementation('junit:junit:4.13.1')
+}
+
+configurations {
+    customTestImpl.extendsFrom testImplementation
+}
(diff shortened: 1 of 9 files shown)


cyrille-artho (approved):

cyrille-artho:
LGTM, but it is not easy to see if all patches have been integrated correctly due to the size of this combined PR. @pparizek , please check also.

pparizek:
I agree with this PR.

Issue #380, opened 2023-08-02
The `VarTracker` listener always throws an `ArrayIndexOutOfBoundsException` exception


samvid25:
Using the `VarTracker` listener results in an `ArrayIndexOutOfBoundsException`. Here is (the smallest possible) example that can cause this:
```java
public class Test {
  public static void main(String[] args) {}
}
```

The error:
```
[WARNING] orphan NativePeer method: jdk.internal.misc.Unsafe.getUnsafe()Lsun/misc/Unsafe;
[SEVERE] JPF exception, terminating: exception during instructionExecuted() notification
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 1
        at gov.nasa.jpf.vm.StackFrame.addOperandAttr(StackFrame.java:533)
        at gov.nasa.jpf.listener.VarTracker.instructionExecuted(VarTracker.java:194)
        at gov.nasa.jpf.vm.VM.notifyInstructionExecuted(VM.java:807)
        at gov.nasa.jpf.vm.ThreadInfo.executeInstruction(ThreadInfo.java:1925)
        at gov.nasa.jpf.vm.ThreadInfo.executeTransition(ThreadInfo.java:1861)
        at gov.nasa.jpf.vm.SystemState.executeNextTransition(SystemState.java:765)
        at gov.nasa.jpf.vm.VM.forward(VM.java:1721)
        at gov.nasa.jpf.search.Search.forward(Search.java:579)
        at gov.nasa.jpf.search.DFSearch.search(DFSearch.java:79)
        at gov.nasa.jpf.JPF.run(JPF.java:613)
        at gov.nasa.jpf.JPF.start(JPF.java:189)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at gov.nasa.jpf.tool.Run.call(Run.java:80)
        at gov.nasa.jpf.tool.RunJPF.main(RunJPF.java:116)
---------------------- JPF error stack trace ---------------------
gov.nasa.jpf.JPFListenerException: exception during instructionExecuted() notification
[...]

PR #428, opened 2023-12-20
Update VarTracker.java

Ayush-Thakur-geek:
FOR issue #380

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ If you are having problems installing and running JPF, please look at the [How
 to install
 JPF](https://github.com/javapathfinder/jpf-core/wiki/How-to-install-JPF) guide.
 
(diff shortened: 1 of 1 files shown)


cyrille-artho:
Thanks, your patch looks good. Could you please try to add a unit test as well so that we have a failing test (from your example) that shows that the patch is needed?

src/classes/modules/java.logging/java/util/logging/FileHandler.java
cyrille-artho, line 1: This change is not necessary; the module name is not part of the package name (but it is reflected in the file name).

cyrille-artho (changes requested):
The new change is unrelated and not needed; please revert it.
Instead, please try to add a unit test that fails if your fix to `VarTracker` is not applied yet.

abimael-turing:
@cyrille-artho If you would like, you can assign this issue to me!

PR #430, opened 2023-12-25
Update FileHandler.java

Ayush-Thakur-geek:
Corrected the package name and imported package for StreamHandler class.


cyrille-artho:
Hi,
Thank you for your patch. The module name is not part of the package name, so the code is correct as is.

PR #431, opened 2024-01-08
fixing issue Orphan NativePeer method with Java 11.0.20.1 #421

### LINKED ISSUE #421: Orphan NativePeer method with Java 11.0.20.1
I'm using the latest version of JPF from the `master` branch, which is [documented](https://github.com/javapathfinder/jpf-core/wiki/System-requirements) as supporting Java SE 11, with Java 11.0.20.1. When invoking JPF, I'm greeted by the following warning.

```
[WARNING] orphan NativePeer method: jdk.internal.misc.Unsafe.getUnsafe()Lsun/misc/Unsafe;
```

The output of `java --version` on my Arch Linux machine is as follows.

```
openjdk 11.0.20.1 2023-08-24
OpenJDK Runtime Environment (build 11.0.20.1+1)
OpenJDK 64-Bit Server VM (build 11.0.20.1+1, mixed mode)
```

Is this something to be concerned about and, if so, which version of Java 11 is recommended?

adysinghh:
fix #421


src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java
```diff
@@ -43,10 +45,10 @@
 public class JPF_jdk_internal_misc_Unsafe extends NativePeer {
 
   @MJI
-  public int getUnsafe____Lsun_misc_Unsafe_2 (MJIEnv env, int clsRef) {
+  public int getUnsafe____Ljdk_internal_misc_Unsafe_2 (MJIEnv env, int clsRef) {
     int objRef = env.getStaticReferenceField("jdk.internal.misc.Unsafe", "theUnsafe");
     return objRef;
-  }
+  }
```
cyrille-artho, line 51: Please try to avoid introducing extra formatting changes, as this may make it more difficult to merge your patch in other branches later.

src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java
```diff
@@ -60,7 +62,7 @@ public void registerNatives____V(MJIEnv env, int clsObjRef) {}
   public int addressSize0____I (MJIEnv env, int objRef) {
       return 0;
   }
-
+
```
cyrille-artho, line 65: Try to remove this change as well.

src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java
```diff
@@ -26,8 +26,10 @@
 import gov.nasa.jpf.vm.NativePeer;
 import gov.nasa.jpf.vm.SystemState;
 import gov.nasa.jpf.vm.ThreadInfo;
+import junit.Test;
```
cyrille-artho, line 29: Are these two imports needed? You do not (yet) add a new test in this patch.

cyrille-artho (changes requested):
Thank you for the update. Please try to minimize the changes; if it is not possible to add a unit test to check against warnings of JPF, then there is no need to add imports for the `junit` test packages.

adysinghh:
@cyrille-artho Hi, I had incorporated all the suggested changes

src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java
```diff
@@ -43,24 +44,21 @@
...
-  public long objectFieldOffset__Ljava_lang_reflect_Field_2__J (MJIEnv env, int unsafeRef, int fieldRef) {
+  public long objectFieldOffset__Ljava_lang_reflect_Field_2__J(MJIEnv env, int unsafeRef, int fieldRef) {
     return fieldOffset__Ljava_lang_reflect_Field_2__I(env, unsafeRef, fieldRef);
   }
 
   @MJI
-  public void registerNatives____V(MJIEnv env, int clsObjRef) {}
-
-  @MJI
-  public int addressSize0____I (MJIEnv env, int objRef) {
```
cyrille-artho, line 60: This function does not seem to be tested by the current unit tests, but are you sure it is not needed at all?

src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java
```diff
@@ -43,24 +44,21 @@
...
+  public int getUnsafe____Ljdk_internal_misc_Unsafe_2(MJIEnv env, int clsRef) {
     int objRef = env.getStaticReferenceField("jdk.internal.misc.Unsafe", "theUnsafe");
-    return objRef;
-  }
+    return objRef; 
+}
 
   @MJI
-  public long objectFieldOffset__Ljava_lang_reflect_Field_2__J (MJIEnv env, int unsafeRef, int fieldRef) {
+  public long objectFieldOffset__Ljava_lang_reflect_Field_2__J(MJIEnv env, int unsafeRef, int fieldRef) {
```
cyrille-artho, line 53: Please add the space before the "(" back so we do not have extra changes in the patch; this is important to make future merges to other branches easier.

cyrille-artho (changes requested):
Thank you for the update. This version is a step forward but still has some changes that do not seem to relate directly to the incorrectly named native peer.
Can you please try to make a minimal patch that does not change any other lines? There are several changes (I've highlighted some of them) that add or remove spaces or empty lines. There is also one change that removes an @MJI method. While the tests still all pass, there may be functionality that the unit tests do not cover but rely on that @MJI method.

adysinghh:
@cyrille-artho Thanks for the valuable feedback! I've incorporated the suggested changes, keeping modifications minimal. Retained necessary methods, removed extra spaces, and ensured consistent spacing.

- Removed unnecessary spaces and empty lines.
- Retained @MJI methods as per feedback.
- Ensured consistent spacing in method signatures.

Please review the changes.

cyrille-artho:
Thanks for adding the code back. Please try to add it back in the same way it was before, as there are still two changes that merely reformat code, which is best avoided (an empty line at the beginning and reordered/reformatted methods at the end).

adysinghh:
@cyrille-artho I had made the changes please review.

src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_Unsafe.java
cyrille-artho: Can you please revert the second change that swaps the two methods, so we have a patch that only has the first (functional) change?

cyrille-artho (changes requested):
Hi,
Please try to isolate the first change. If you are not familiar with "diff" and "patch", you can try to apply the first change again manually on a new branch and make a new PR.
Thank you for your perseverance!

shreyasCs012:
the native peers method present in the `\src\peers\gov\nasa\jpf\vm` had a  "mangled" which was old so it has been corrected to the new location which is `\jdk\internal\misc`

cyrille-artho:
Subsumed by PR #492

Issue #250, opened 2020-07-31
Missing check for STATIC method when executing


ilovejpf:
Hi! I find that jpf is missing check `static` method, which can be severe! Let me show you an example.
First, you write a class C, which will later be compiled to a method call to use INVOKESTATIC.
```
cat > C.java <<EOF
class C {
   public static void main(String[] a) {
     int i = 123;
     D.m("foobar");
     System.out.println(i);
   }
}
EOF
```
Then you write the static method `m` in class D.
```
cat > D.java <<EOF
class D {
   public static String m(String s) { return s; }
}
EOF
```
Then you can run `javac C.java D.java` to compile, and both `java C` and `java -jar build/RunJPF.jar +classpath=. C` should work fine.

Then you remove that `static` in class D by running `sed -i s/static// D.java; javac D.java`, then `java C` will throw `IncompatibleClassChangeError` as expected.
```
java.lang.IncompatibleClassChangeError: Expected static method D.m(Ljava/lang/String;)Ljava/lang/String;
	at C.main(C.java:4)
```
However, `java -jar build/RunJPF.jar +classpath=. C` will print out `248` (please note that this number should be `123`! ) and end up with `no errors detected`. Missing the check can lead to executing something unexpected and destroy certain variables in the program.
The fix may be adding a check for whether the method to be invoked is actually static or not when executing INVOKESTATIC.
Thanks!

cyrille-artho:
Hi,
Thank you very much for noticing this. This is an issue that we hope won't be too hard to fix, but it may be difficult to test.
We would need the following:
1. C.java and D.java, which are both compiled as usual during gradle's `testCompile` phase.
2. Before the tests run, we'd need a task that (a) creates a modified version of D.java in a temporary directory, (b) compiles it, and (c) moves D.class (the compiled file) to the location of the other compiled files. 
3. Now the tests can run.

This special task in (2) can be implemented using doFirst:
https://docs.gradle.org/current/userguide/build_lifecycle.html

The actions inside that task could be implemented as shell tasks, as in:
https://discuss.gradle.org/t/run-a-process-before-tests/25690/2

However, this would make this particular gradle task less portable and more failure-prone; it should at the very least not prevent the other tests from running if it cannot execute. Perhaps it's also possible to use a gradle-native task to compile the modified D.java.

What do you think?

alokprince:
can I try to solve this issue?
I am a beginner and I am willing to learn and work on this issue as my first issue.

[...]

PR #438, opened 2024-03-19
Added Static method check before executing

CrypticMessenger:
in ref to issue: https://github.com/javapathfinder/jpf-core/issues/250#issue-669914302
- Added isStatic() class check before executing in ```INVOKESTATIC.java```


cyrille-artho:
Hi,
Thank you for the patch.
Please try to add a test that fails without the patch but succeeds with it, so we can see what use case gets fixed with the patch.

cyrille-artho (changes requested):
Please add a unit test that fails on the previous version of JPF but succeeds with your patch.

Issue #439, opened 2024-03-21
`String.equals(Object)` makes JPF on Java 11 crash


quadhier:
Hi @cyrille-artho , I found a bug in implementation of `String.equals(Object)` in JPF and it only exists on `master` branch (for Java 11). A simple program like the following one would reveal this issue.

```
public class StringCmp {
  public static void main (String[] args) {
    System.out.println("a".equals(new Object()));
  }
}
```

The fix is straightforward, yet it requires some knowledge of native peers. I think it would be a good candidate for "good first issue". If you agree, perhaps you could tag this issue with the label, allowing new contributors the opportunity to resolve the bug.

Dashlet26:
To fix the bug that only exists on the master branch when using String.equals(Object), alter the code in the StringCmp class to correctly compare a String with an Object.

cyrille-artho:
Hi,
Thank you for finding the bug. In this case, adding a test and fix is indeed easy (as you have already provided the test and adding a test merely involves using the skeleton from https://github.com/javapathfinder/jpf-core/wiki/Writing-JPF-tests).
This fix involves a small change in the native peer.
I'll mark it as a good issue for beginners.

Dashlet26:
Hi , @quadhier 
I known how to solve this type of problem but unable to get the file which you are talking about .

Harshal-Rembhotkar:
Hi @cyrille-artho ,
i know the problem's solution but unable to get file which of this bug #439 , please help me..

cyrille-artho:
In this case, `String` needs a new `equals` method. You can find the model class for `String` in `src/classes`.

Harshal-Rembhotkar:
Thanks @cyrille-artho...

Harshal-Rembhotkar:
Hello everyone , i am unable to build a jpf-core project locally .
 issue: 
["private static JavaUtilJarAccess javaUtilJarAccess;
                 ^
  symbol:   class JavaUtilJarAccess
  location: class SharedSecrets" ]
i facing above issue ,please help me ...
![Screenshot 2024-03-27 214543](https://github.com/javapathfinder/jpf-core/assets/147272231/0018863c-a3f5-4ba8-b191-8c553f1d64fd)

![Screenshot 2024-03-27 213943](https://github.com/javapathfinder/jpf-core/assets/147272231/72325fa0-6710-4b24-b286-a8043278b687)

abimael-turing:
[...]

PR #440, opened 2024-03-29
for #439

Ayush-Thakur-geek:
Added a check for checking whether the object being compared is a string.


cyrille-artho:
Hi,
Your change makes eight unit tests fail. Can you please check? Try `./gradlew clean test`.

Ayush-Thakur-geek:
Actually I am not able to get the project properly setup in my machine and this is due to missing of some  jdk.internal.misc.
Could you please tell me how to resolve this.

cyrille-artho:
You have to make sure that you use OpenJDK version 11. Older versions of Java don't have modules, and newer versions of Java have other incompatibilities. Oracle's JDK is also slightly different from OpenJDK.

PR #446, opened 2024-04-25
Unsafe fixing new branch

eklaDFF:
Fixed the Unsafe class


cyrille-artho:
Can you please rebase your patch on `javapathfinder:java-17`? That will remove the ten commits that are already merged.
I have now fixed the configuration issue in the workflow so the CI build will be attempted when you update the patch.

cyrille-artho:
I'm sorry that it is a bit complicated to set up the process properly, but once we have this down, subsequent PRs are going to be much more efficient.

Issue #457, opened 2024-05-31
`java.net.URI.toURI` broken


cyrille-artho:
The model class implements `toURI`, but this function was not tested with a unit test. We therefore need a unit test and then a fix.

1. Create a unit test that constructs a simple `URI` from a `File` with a `String` that has the right URI syntax.

2. Look into ways to obtain a working URI by fixing a dependency in the JPF model classes, most likely by adding `setJavaNetURIAccess`, which is probably going to resemble `setJavaNetURLAccess`.

Also look into the class description JavaDoc of `src/classes/modules/java.base/jdk/internal/misc/SharedSecrets`, which still refers to Java 1.5/Java 6 and is therefore seriously outdated.

If the model class for `SharedSecrets` can indeed be removed, we want a careful review of its use of `Unsafe` and whether removal might cause problems later.

PR #460, opened 2024-06-03
java.net.URI.toURI broken

Harsh4902:
- [x] Unit test for 'toURI' method
- [ ] Obtain working URI by fixing a dependency in the JPF model classes

- fixing issue #457


Harsh4902:
@cyrille-artho please review this PR.

src/tests/gov/nasa/jpf/test/java/io/FileTest.java
```diff
@@ -102,5 +104,19 @@ public void testEquals() {
...
+    File file = new File("testfile.txt");
+    URI expectedURI = null;
+    try {
+      expectedURI = new URI("file:testfile.txt");
+    } catch (URISyntaxException e) {
+      fail("URISyntaxException thrown while constructing expected URI");
+    }
+
+    URI actualURI = file.toURI();
+    assertNotNull("URI should not be null",actualURI);
```
cyrille-artho, line 119: The non-`null` check is subsumed by the assertion below, so you can remove this line.

cyrille-artho (changes requested):
Good job!
As this test will fail without a fix in jpf-core, let's wait with merging the PR until Part 2 (the fix) is also ready.

src/tests/gov/nasa/jpf/test/java/io/FileTest.java
Harsh4902: Okay will remove it.

cyrille-artho:
Thanks for the update. This test is a good way to verify an upcoming fix of `toURI`. Once you have a fix for `toURI`, we can merge the PR.

Harsh4902:
@cyrille-artho what should be the standard output of `toURI` method based on our use case in modbat?

cyrille-artho:
We always use the regular JVM as a baseline. In this case, your test may fail because `toURI` seems to include the absolute path in the filename. Try to adapt the test case so that the assertion holds when you execute it on the regular JVM (without JPF), as we want the behavior under JPF to be equivalent.

output.txt
```diff
@@ -0,0 +1 @@
+After.
```
cyrille-artho, line 1: Can you remove this file for the PR?

cyrille-artho (changes requested):
You seem to have accidentally included an extra file.

src/tests/gov/nasa/jpf/test/java/io/FileTest.java
```diff
@@ -102,5 +104,18 @@ public void testEquals() {
...
   }
-  
+
+  @Test
+  public void testToURI(){
+    File file = new File("testfile.txt");
+    URI expectedURI = null;
+    try {
+      expectedURI = new File(file.getAbsolutePath()).toURI();
+    } catch (Exception e) {
```
cyrille-artho, line 114: Catch a specific subtype of `Exception` so that general failures do not cause the test to pass.

cyrille-artho:
Strange that the test passes while a simple program using `toURI` fails with JPF:
With this being `Test.java`:

```
import java.io.*;

public class Test {
  public final static void main(String[] args) throws Exception {
    File f = new File("test");
    System.out.println(f.toURI().toURL());
  }
```

I have run this with `jpf +classpath=. Test` and get the following:

```
java.lang.NoSuchMethodException: jdk.internal.misc.SharedSecrets.setJavaNetUriAccess(Ljdk/internal/misc/JavaNetUriAccess;)V
	at java.net.URI.<clinit>(URI.java:3626)
	at java.io.File.toURI(File.java:131)
	at Test.main(Test.java:6)
```

cyrille-artho:
Make sure the tests run on JPF, not the host JVM:
https://github.com/javapathfinder/jpf-core/wiki/Writing-JPF-tests

src/tests/gov/nasa/jpf/test/java/io/FileTest.java
Harsh4902: Updated unit test

output.txt
Harsh4902: Removed it.

Harsh4902:
@cyrille-artho Please review it. Hopefully its resolved now.

cyrille-artho (approved):

cyrille-artho:
Thanks, looks good!

PR #465, opened 2024-06-22
shoaib/issue#439

sshaikshoaib:
i have changed .equals function to .compareTo function. hope it works

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.debug.settings.onBuildFailureProceed": true
+}
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_String.java
@@ -172,7 +172,7 @@ public class JPF_java_lang_String extends NativePeer {
     String s1 = env.getStringObject(objRef);
     String s2 = env.getStringObject(argRef);
 
-    return s1.equals(s2);
+    return s1.compareTo(s2)==0;
   }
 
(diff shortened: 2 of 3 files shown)


cyrille-artho:
Thank you for your implementation. Please also provide a unit test so we know that the new version of the code works as intended. See the wiki for instructions on how to write unit tests: https://github.com/javapathfinder/jpf-core/wiki/Writing-JPF-tests

cyrille-artho (changes requested):
Please add a unit test, as this part of the code is currently not covered by a test, so we don't know if your change fixes the issue.

sshaikshoaib:
Hi @cyrille-artho I am a beginner and I have a good knowledge on java.so i used this knowledge to rectify the issue.
so give me some time as Iam a beginner I don't know how to write test cases.
But i have an Immense interest in contributing to java projects.so can you just me how to write test cases

sshaikshoaib:
i have added the required tests for the updated code. please revies that and suggest me if any updates required

cyrille-artho:
Hi,
Thank you for the test. It fails, which means the test shows the problem with the typecast not happening as expected, but the code change in the main code does not fix the bug yet.

cyrille-artho:
Thanks! This patch seems to fix the problem. To merge it, can you please remove the extra file `.vscode/settings.json`? You accidentally added it, and you can simply use `git rm -f .vscode/settings.json` followed by a commit on your branch (shoaib/issue#439) to remove it. Once you push that change, this pull request will be updated automatically.

cyrille-artho:
Subsumed by #491

PR #466, opened 2024-06-26
`Package.toString()` not working in JPF

Harsh4902:
- I have created a test case in `URLClassLoaderTest` which is failing because `Pcakage.toString()` method is not working and throwing `NullPointerException`

### DIFF
--- a/src/tests/gov/nasa/jpf/test/java/net/URLClassLoaderTest.java
+++ b/src/tests/gov/nasa/jpf/test/java/net/URLClassLoaderTest.java
@@ -519,4 +519,17 @@ public class URLClassLoaderTest extends LoadUtility {
     }
     movePkgBack();
   }
+
+  @Test
+  public void getPackageInfo() throws MalformedURLException {
+    movePkgOut();
+     if(verifyNoPropertyViolation())  {
+       URL[] urls = { new URL(dirUrl) };
+       TestClassLoader cl = new TestClassLoader(urls);
+         assertNotNull(cl.getDefinedPackage("java.lang"));
+         assertNotNull(cl.getDefinedPackage("java.lang").getName());
+         assertNotNull(cl.getDefinedPackage("java.lang").toString());
+    }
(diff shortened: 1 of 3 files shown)


src/tests/gov/nasa/jpf/test/java/net/URLClassLoaderTest.java
```diff
@@ -519,4 +519,17 @@ public void testThrownException() throws ClassNotFoundException, MalformedURLExc
...
   }
+
+  @Test
+  public void getPackageInfo() throws MalformedURLException {
+    movePkgOut();
+     if(verifyNoPropertyViolation())  {
+       URL[] urls = { new URL(dirUrl) };
+       TestClassLoader cl = new TestClassLoader(urls);
+         assertNotNull(cl.getDefinedPackage("java.lang"));
+         assertNotNull(cl.getDefinedPackage("java.lang").getName());
```
cyrille-artho, line 530: change to `assertEquals` with the package name as the expected value

Harsh4902: changed.

cyrille-artho:
Thanks. For the main branch, we wait with merging code until the test passes.

Harsh4902:
@cyrille-artho @pparizek I have changed the approach of loading package. We can access package by using object of `Class` class for current class. With this approach, we can further call Package.toString() method, which will not fail on JPF.

cyrille-artho:
Great! Can you please try to make the patch (in terms of changes) a bit smaller? For example, it's usually better to explicitly import the classes you need rather than using the wildcard, so the change from `import java.util.A; import java.util.B; ...` to `import java.util.*;` is not ideal. It also makes the patch bigger, which will make it harder to merge it into other branches.
Please check if you can do that (and push the changes again to this branch, which will auto-update the PR).

cyrille-artho:
Thanks!

PR #467, opened 2024-06-28
add Logger interface and getLogger() method

eklaDFF:
1. Add `Logger` interface
2. add `getLogger()` method


cyrille-artho:
Thanks!

PR #476, opened 2024-07-12
Terminate the thread when unhandled exception happens.

aoli-al:
JPF is failing the added test `ThreadExceptionTest` because it does not terminate a thread when an unhandled exception occurs. This PR fixed this issue by continuing to execute the `DIRECTCALLRETURN` instruction instead of throwing the UncaughtException.


cyrille-artho:
Thanks, this test indeed fails with JPF; even though the exception occurs, the thread state is not handled correctly, as you noted.

PR #478, opened 2024-07-21
Tests and MethodHandle for Instance and Static Field

eklaDFF:
We have added 3 Tests to test instance field and 3 for static field.....


cyrille-artho:
Looks good, thank you very much!

PR #479, opened 2024-07-21
BufferTest Codes and Changes

eklaDFF:


src/classes/modules/java.base/java/lang/String.java
```diff
@@ -595,13 +595,34 @@ public char charAt(int index) {
...
 	native public int offsetByCodePoints(int index, int codePointOffset);
-	native public void getChars(int srcBegin, int srcEnd, char dst[], int dstBegin);
+
+
+	/*
+	In OpenJDK17, this method (getChars(int srcBegin, int srcEnd, char dst[], int dstBegin)) is not a native method.
+	So commented this out and it's corresponding peer method also.
+	Below is its new implementation
+	*/
+//	native public void getChars(int srcBegin, int srcEnd, char dst[], int dstBegin);
```
cyrille-artho, line 605: remove instead of commenting out

src/classes/modules/java.base/java/lang/String.java
```diff
@@ -595,13 +595,34 @@ public char charAt(int index) {
...
+		}
+	}
+
+
 	native void getChars(char dst[], int dstBegin);
 
 	@Deprecated
 	native public void getBytes(int srcBegin, int srcEnd, byte dst[], int dstBegin);
 	public void getBytes(byte[] dst, int srcPos, int dstBegin, byte coder, int length){
-		getBytes(srcPos,(srcPos+length),dst,dstBegin);
```
cyrille-artho, line 604: remove instead of commenting out

cyrille-artho (changes requested):
Looks good! Furthermore, the unit tests all pass now, which is great!
All I can see are two minor things: We want to use git to handle changes, so we want to avoid commenting out old code and instead remove it. This makes it easier to maintain the code in the long term.
Please update the PR accordingly, and I will merge it.

eklaDFF:
@cyrille-artho 
Made those changes

src/classes/modules/java.base/java/lang/String.java
eklaDFF: Made those changes

cyrille-artho (approved):

PR #488, opened 2024-08-01
replace deprecated Constructor based initialization of primitive type…

eklaDFF:
…'s Wrapper Class with WrapperClass.valueOf(value)

### DIFF
--- a/src/tests/gov/nasa/jpf/test/mc/basic/AttrsTest.java
+++ b/src/tests/gov/nasa/jpf/test/mc/basic/AttrsTest.java
@@ -361,7 +361,7 @@ public class AttrsTest extends TestJPF {
     }
   }
   
-  @SuppressWarnings("deprecation")
+  @SuppressWarnings("removal") // Integer o = new Integer(v)
   @Test public void testInteger() {
     if (verifyNoPropertyViolation()) {
       int v = 42;
--- a/src/tests/gov/nasa/jpf/test/java/lang/BoxObjectCacheTest.java
+++ b/src/tests/gov/nasa/jpf/test/java/lang/BoxObjectCacheTest.java
@@ -17,12 +17,11 @@
  */
 package gov.nasa.jpf.test.java.lang;
 
+import gov.nasa.jpf.util.test.TestJPF;
+import org.junit.Test;
+
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-
-import gov.nasa.jpf.util.test.TestJPF;
-
-import org.junit.Test;
 
 /**
  * regression test for java.lang.Integer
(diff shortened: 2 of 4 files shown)


src/tests/gov/nasa/jpf/test/mc/basic/AttrsTest.java
```diff
@@ -360,35 +360,50 @@ int goModel (double d, int i) {
...
+  // Example : { Integer o = new Integer(v) } works but { Integer o = Integer.valueOf(v) } don't.
+
+//  @SuppressWarnings("deprecation")
+//  @Test public void testInteger() {
+//    if (verifyNoPropertyViolation()) {
+//      int v = 42;
+//      Verify.setLocalAttribute("v", 4200);
+//
+//      // explicit
+//      Integer o = Integer.valueOf(v);
```
cyrille-artho, line 379: I thought that the first half of the test still works? If we can keep the part with the explicit cast, that would be preferable.

src/tests/gov/nasa/jpf/test/mc/basic/AttrsTest.java
```diff
@@ -360,35 +360,50 @@ int goModel (double d, int i) {
...
+//      int j = o.intValue();
+//      int attr = Verify.getLocalAttribute("j");
+//      assert attr == 4200;
+//
+//      // semi autoboxed
```
cyrille-artho, line 384: If the part above works, we can start comment out this part.
I think it is nicer to use `/*` here, akin to the commented-out part below. Definitely not `//` in addition to `/*`.
With `/*`...`/*`, we clearly mark two blocks of code as being parts of the test that do not (or do no longer) work. With `//`, we lose that information.

cyrille-artho (changes requested):

src/tests/gov/nasa/jpf/test/mc/basic/AttrsTest.java
eklaDFF: Explicit means `Integer in = new Integer ()` ?

src/tests/gov/nasa/jpf/test/mc/basic/AttrsTest.java
eklaDFF: Actually, none of the Assertion works.

Pavel taught me that `Local Variable` and `Local Attribute` are different entities in JPF.
Constructor type initialisation passes the value of `Local Variables` to `Local Attributes` but in case of new method it doesn't.

eklaDFF:
There are two assert statements and none of them work. Pavel suggested that
this might be due to Local Variables and Local Attributes miscommunication
in new method.

cyrille-artho:
I see. In this case, please modify this PR to first retain the problematic test as is (with no code changes, keeping the code that produces compiler warnings).
I think it might be better to suppress the remaining compiler warnings in that problematic test in a second PR, to avoid commenting out or removing a lot of code.

eklaDFF:
Made those changes, please review

cyrille-artho (approved):

cyrille-artho:
Great, thanks!

PR #491, opened 2024-08-13
 Fix and Test String.equals(Object) Implementation in JPF

shreyasCs012:
This pull request addresses a bug in the Java Pathfinder(JPF) related to implementing the 'String. equals(Object)' method. I have made the following changes :

### Test Addition :

   I added a test to check if the equals method works if a non-String object was passed as an argument but the previous implementation of the equals method did not check this condition which was revealed via test.

### Bug Fix :

1.   argRef Handling:

     -   The argRef variable which is one of the parameters passed in 'equals__Ljava_lang_Object_2__Z '. The purpose of the argRef 
          is a reference to the symbolic object in the jpf heap but it does not check if the object is a String object before comparing 
          the objects(concrete).     

2. Updates : 

     -  The method now uses env.isInstanceOf(argRef, "java.lang.String") which checks if the object is a String before it checks the 
          equality of the different objects.


cyrille-artho:
Thanks!

PR #497, opened 2024-09-13
VM.getSavedProperty("os.name") not supported 

### LINKED ISSUE #495: VM.getSavedProperty("os.name") not supported
Some properties have to be set at the time when the VM is initialized; this is not the case for `os.name`. If a class later calls `VM.getSavedProperty("os.name")`, the result is an `IllegalArgumentException` because the property has not been set.

Harsh4902:
### Description
This PR adds `os.name` key into native peer for `java.lang.System` to fethc OS name from Host VM. Also adds test case to test `VM.getSavedProperty` functionality in JPF.

### Related issue(s)
Fixes #495


Harsh4902:
@cyrille-artho I am not sure why `gov.nasa.jpf.vm.multiProcess.*` tests are failing. Working on it.

Issue #429, opened 2023-12-21
JPF fails with java.lang.IllegalArgumentException: Null charset name


aoli-al:
Hi, I'm trying to use JPF to run the following simple program

```
import java.nio.file.Path;
public class BoundedBuffer {
  public static void main(String [] args) {
    Path.of("/tmp");
  }
}
```

I got the following exception:

```
JavaPathfinder core system v8.0 (rev 1a704e1d6c3d92178504f8cdfe57b068b4e22d9c) - (C) 2005-2014 United States Government. All rights reserved.


====================================================== system under test
BoundedBuffer.main()

====================================================== search started: 12/21/23, 3:01 PM
[WARNING] orphan NativePeer method: jdk.internal.misc.Unsafe.getUnsafe()Lsun/misc/Unsafe;

====================================================== error 1
gov.nasa.jpf.vm.NoUncaughtExceptionsProperty
java.lang.IllegalArgumentException: Null charset name
	at java.nio.charset.Charset.lookup(Charset.java:455)
	at java.nio.charset.Charset.forName(Charset.java:526)
	at sun.nio.fs.Util.<clinit>(Util.java:40)
	at sun.nio.fs.UnixFileSystem.<init>(UnixFileSystem.java:51)
	at sun.nio.fs.LinuxFileSystem.<init>(LinuxFileSystem.java:39)
	at sun.nio.fs.LinuxFileSystemProvider.newFileSystem(LinuxFileSystemProvider.java:46)
	at sun.nio.fs.LinuxFileSystemProvider.newFileSystem(LinuxFileSystemProvider.java:39)
	at sun.nio.fs.UnixFileSystemProvider.<init>(UnixFileSystemProvider.java:56)
	at sun.nio.fs.LinuxFileSystemProvider.<init>(LinuxFileSystemProvider.java:41)
	at sun.nio.fs.DefaultFileSystemProvider.<clinit>(DefaultFileSystemProvider.java:35)
	at java.nio.file.FileSystems.getDefault(FileSystems.java:185)
	at java.nio.file.Path.of(Path.java:147)
	at BoundedBuffer.main(BoundedBuffer.java:121)


====================================================== snapshot #1
thread java.lang.Thread:{id:0,name:main,status:RUNNING,priority:5,isDaemon:false,lockCount:0,suspendCount:0}
  owned locks:java.lang.Class@20b,java.lang.Class@278
  call stack:
	at sun.nio.fs.Util.<clinit>(Util.java:40)
[...]

PR #506, opened 2024-11-17
Added model classes for java.nio.file and java.nio.charset packages #429

joalen:
Turns out, there needs to be new model classes found in the java.nio.charset and java.nio.file packages. I believe I made them as small as I could to patch the issue with the original reproducer. 

The following happens when this new solution works:
```
JavaPathfinder core system v8.0 (rev 8fe399b8718b1a25055f18b61707ca0cf03112f0) - (C) 2005-2014 United States Government. All rights reserved.


BoundedBuffer.main()

====================================================== search started: 11/16/24, 10:42 PM

no errors detected

====================================================== statistics
elapsed time:       00:00:00
states:             new=1,visited=0,backtracked=1,end=1
search:             maxDepth=1,constraints=0
choice generators:  thread=1 (signal=0,lock=1,sharedRef=0,threadApi=0,reschedule=0), data=0
heap:               new=555,released=35,maxLive=0,gcCycles=1
instructions:       9557
max memory:         2056MB
loaded code:        classes=93,methods=2360

====================================================== search finished: 11/16/24, 10:42 PM```


joalen:
@cyrille-artho 

Yep, seems like there were some redundant classes and Charset caused issues with the current JPF pipeline. I created a new commit to address some of the issues and it should now work. Thank you again for catching that!!

cyrille-artho:
Thanks, with this, we have a few classes less to maintain, and the tests pass.

PR #510, opened 2025-01-09
JPF lacks a check on whether a statically invoked method is indeed static

sanidhya00081:
Integrated a mechanism to detect and prevent invalid INVOKESTATIC calls by validating method properties during runtime in JPF (line 274).
Optimized the test pipeline by using Gradle-native methods to manage input-output dependencies, ensuring task portability and proper execution order.


build.gradle
```diff
@@ -269,6 +271,33 @@ tasks.register('buildJars') {
     dependsOn createRunJpfJar
     dependsOn createRunTestJar
 }
+task.register("modifyAndCompileD") {
```
cyrille-artho, line 274: Should probably be `tasks.register`?

```
> Could not get unknown property 'task' for root project 'jpf-core' of type org.gradle.api.Project.
```

cyrille-artho (changes requested):
I think there is a typo, please fix it so we can try the CI build/test again.

sanidhya00081:
yeah sure @cyrille-artho

cyrille-artho:
Thanks! The build file works now, but one of the paths seems to be wrong, and `D.java` is missing in your PR, so it can't be found:
```
 - Property '$1' specifies file '/home/runner/work/jpf-core/jpf-core/src/main/java/D.java' which doesn't exist.
 ```
See the full log here: https://github.com/javapathfinder/jpf-core/actions/runs/12714274234/job/35472592178?pr=510

**Comment by sanidhya00081** [comment]
Thanks for quoting the issue @cyrille-artho . I would be working on the on some amendments and would come up with a new PR.

**Comment by sanidhya00081** [comment]
Hey! @cyrille-artho could this issue be assigned to me?

**Comment by cyrille-artho** [comment]
> Hey! @cyrille-artho could this issue be assigned to me?

This is a PR, not an issue. I can assign an issue to you once you have made a successful PR.

This version works better than the previous one, but it still results in build failures:

```
  - Property '$1' specifies file '/home/runner/work/jpf-core/jpf-core/src/main/java/D.java' which doesn't exist.
    
    Reason: An input file was expected to be present but it doesn't exist.
    
    Possible solutions:
      1. Make sure the file exists before the task is called.
      2. Make sure that the task which produces the file is declared as an input.
    
    For more information, please refer to https://docs.gradle.org/8.4/userguide/validation_problems.html#input_file_does_not_exist in the Gradle documentation.
  - Gradle detected a problem with the following location: '/home/runner/work/jpf-core/jpf-core/build/classes/java/main'.
    
    Reason: Task ':createJpfClassesJar' uses this output of task ':modifyAndCompileD' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':modifyAndCompileD' as an input of ':createJpfClassesJar'.
      2. Declare an explicit dependency on ':modifyAndCompileD' from ':createJpfClassesJar' using Task#dependsOn.
      3. Declare an explicit dependency on ':modifyAndCompileD' from ':createJpfClassesJar' using Task#mustRunAfter.
```

I can see that `jpf-core` appears twice in a row in the path above (`jpf-core/jpf-core`), which is probably the reason the file is not found.
Gradle build dependency rules have changed a bit recently, so try to see if explicit dependency declarations fix the build.
Make sure that the Gradle build is successful on your system, as this will accelerate the CI/review cycle.

**Comment by sanidhya00081** [comment]
The build was successful on my system @cyrille-artho 

Screenshot 2025-01-13 004539

**Comment by cyrille-artho** [comment]
Great! I can confirm that the new build now works. This resolves the most difficult step.
To completen the work on issue #250, we need two more things:
1. A test case that demonstrates that JPF incorrectly allows a static call to a non-static method to be executed, even though the "static" keyword had been removed and the file recompiled. You can find out how to write tests to run inside JPF on the wiki: https://github.com/javapathfinder/jpf-core/wiki/Writing-JPF-tests 
2. A fix that adds an additional check for the "static" flag in the method and correctly rejects such a call.

**Comment by sanidhya00081** [comment]
@cyrille-artho Thanks!! I would surely be working on these two points and will revert back in a day.

**Comment by sanidhya00081** [comment]
I have created the test file and have successfully checked whether JPF incorrectly allows a static call to a non-static method. The build was successful. @cyrille-artho

**Comment by cyrille-artho** [comment]
Hi,
Thank you for your first attempt at a test. Note that just because an instance is `null`, the call is still a dynamic call. Your test therefore does not test the right thing.
A call to a static method is generated by using the class name instead of an instance. Your example `C.java` does this. If you run `C.main` against a re-compiled `D.java` (where the keyword `static` has been removed), you get a very different exception:
```
Exception in thread "main" java.lang.IncompatibleClassChangeError: Expected static method 
```
So your test can try that and call `C.main` instead; `C.main` then calls `D.m`, and we get the exception. (You can also call `D.m` directly from your test, as your regular unit tests will be compiled against the original static version of D.m. So we don't need to keep `C.java` anymore.)

If you adapt the test, it should first *fail*, because JPF does not check the `static` flag properly at run time, so the call to `D.m` succeeds even if `D` has been recompiled with a dynamic method.

**Comment by cyrille-artho** [comment]
I think it's a good idea to first try to get a "properly failing" test, so we are sure that the expected exception is checked against in the right way. After that, the one-line change in JPF to fix the test can easily be confirmed as valid by rerunning the test against the fix.

**Comment by sanidhya00081** [comment]
@cyrille-artho  Sure!! I would that and will revert back asap. I have understood the issue. Thank you @cyrille-artho for the clear guidance. It really makes easy for me to proceed towards next steps.

**Comment by sanidhya00081** [comment]
I have tested the failure case successfully and have made a PR to show it. But I am not able to get what can I do next. @cyrille-artho  can you give me some suggestions.

File: __test__ [hunk 2]
```diff
@@ -0,0 +1 @@
+***
```
cyrille-artho, line 1: Please remove any temporary files in the PR. Change the test so any such files are removed at the end of the test.

src/main/gov/nasa/jpf/D.java
```diff
@@ -0,0 +1,8 @@
+package gov.nasa.jpf;
```
cyrille-artho, line 1: These files should be in `src/test`.

src/main/java/C.java
```diff
@@ -0,0 +1,7 @@
+class C {
```
cyrille-artho, line 1: Move to `src/test`.

src/main/java/D.java
```diff
@@ -0,0 +1,3 @@
+class D {
```
cyrille-artho, line 1: Move to `src/test`.

cyrille-artho (changes requested):
Please remove temp. files and move the test files to the right place. You might not need `C.java` anymore. Furthermore, the tests no longer compile, so please check your build.

cyrille-artho:
As of now, some test files need to be moved to `src/test`. `D.java` is not found in one case during compilation, causing it to fail. Please check:
```
/home/runner/work/jpf-core/jpf-core/src/tests/gov/nasa/jpf/StaticCallToNonStaticTest.java:13: error: cannot find symbol
                D.m(); // This should throw IncompatibleClassChangeError after D.java is recompiled without static
                ^
  symbol:   variable D
  location: class StaticCallToNonStaticTest
```

sanidhya00081:
@cyrille-artho  I have made the suggested changes for cleaning up files and removing them at the end. Made some dependency changes in build.gradle. Updated StaticCallToNonStaticTest.java.

src/main/gov/nasa/jpf/D.java
cyrille-artho: Please remove this file (as it is no longer needed, we have `src/test/.../D.java` now.

src/tests/gov/nasa/jpf/StaticCallToNonStaticTest.java
```diff
@@ -0,0 +1,42 @@
...
+public class StaticCallToNonStaticTest extends TestJPF {
+
+    @Test
+    public void testStaticCallToNonStatic() {
+        // Create an instance of D
+        D instance = new D();
+    
+        // Expect IncompatibleClassChangeError when calling instance.m()
+        try {
+            System.out.println("Calling m() method");
```
cyrille-artho, line 18: The `println` statement here can be deleted, as it is not desirable in a unit test to print to the console.

cyrille-artho:
Thank you very much! The process is mostly working now, but you accidentally changed the static to a dynamic call in the test; please see my comments.

Another thing I have not commented on, for some reason, this time, only one test (for your new feature) was run, rather than over 1000 tests. I don't know if this is due to change in the build/test process in `build.gradle` or due to caching by GitHub.

build.gradle
```diff
@@ -76,16 +114,20 @@ tasks.named('compilePeersJava') {
 tasks.named('compileTestJava') {
     dependsOn(copyLibs)
     dependsOn(copyResources)
+    mustRunAfter(modifyAndCompileD)
```
cyrille-artho, line 117: I think this is reversed. I am not very familiar with the Gradle DSL, and the documentation is not 100 % clear, but this dependency and others (even some that have been declared previously) might be redundant or reversed. Please look at the dependencies and ensure that the original unit tests are preserved, so we have again over 1000 tests in the CI build/test.
You can check this by running `./gradlew clean test`.

cyrille-artho (changes requested):
Regarding build dependencies.

sanidhya00081:
@cyrille-artho  If the method `m()` is intended to be non-static, we need to create an instance of the `D` class

sanidhya00081:
Hey @cyrille-artho I have made the suggested changes. I think now the issues is solved and the build was also successful. And the dependency issue is solved I think.

cyrille-artho:
Please try the build/test on your system before pushing:
Your new `src/test/D.java` is in the wrong place; the directory should be the same as the package name (`src/test/pkg/name/D.java`).
You also deleted two existing test files (`StringConcatenationTest.java` and another one); please reinstate them.

sanidhya00081:
Hey @cyrille-artho could you please confirm whether my changes are visible on the master for review. I am facing some issues with the internet, so the it is taking time to load.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
cyrille-artho, line 1: Please remove any files with local settings (editor settings etc.) from the repo. You can keep a local copy and add `.vscode/settings.json` to `.gitignore`.

cyrille-artho (changes requested):
Your code changes look good, but you still have the accidental deletion of two test files `src/tests/java11/StringConcatenationTest.java` and `src/tests/java11/StringEqualsTest.java`. Please add them back, and conversely, remove your local settings (`settings.json`).
The main test suite still does not run, please check.

cyrille-artho:
Yes, I can see your changes now. While the test outcome is reported as successful, only one unit test is actually run: `Summary: 1 tests, 1 passed, 0 failed, 0 skipped`. Please check on your system if this is also the case for you.

sanidhya00081:
@cyrille-artho  After the updates suggested, I am also getting the same message `Summary: 1 tests, 1 passed, 0 failed, 0. skipped`. I am not able to understand why this is happening or why this message is occurring.

cyrille-artho:
It is not obvious to me; most likely, the problem is in the changes to `build.gradle`.

Perhaps, your additional directive `sourceDir` is too specific when you list `src/main/java`, as it does not match the file layout in `jpf-core`. Still, that does not explain that compilation seems to succeed.

I recommend the following: Clone `jpf-core` again from your forked repo, into a separate location that you can use as a "starting point" or reference. Start again from the original `build.gradle` while keeping all your other file changes. Try to adapt *one* of your changes to `build.gradle` from your PR at a time. You probably have to combine several chunks at a time sometimes, so the changes are syntactically and semantically consistent.

Then, your new test will probably not work correctly (because it requires the new build rules), but at least initially, you will have 1010 unit tests again:

`Summary: 1010 tests, 1010 passed, 0 failed, 0 skipped`

If a particular change results in this becoming one unit test, then you know what feature causes problems.

After having identified the location of the problem, it will be much easier to fix it.

build.gradle
```diff
@@ -76,16 +112,25 @@ tasks.named('compilePeersJava') {
 tasks.named('compileTestJava') {
     dependsOn(copyLibs)
     dependsOn(copyResources)
+    shouldRunAfter(modifyAndCompileD)
```
cyrille-artho, line 115: This dependency should be the other way around: in the declaration of `modifyAndCompileD`, that task should run after `compileTestJava`, so we first compile the tests normally and then modify `D.java` and re-compile it.

cyrille-artho:
Thanks for tidying up the PR. The 1010 basic tests now run and pass, but the new test does not run for some reason.
One thing you could try is to add a new Gradle target specifically for your new test and run this as a third test-related task in `.github/workflows/simple_build.yml`:
```
    - name: Run test on modified class file
      run: ./gradlew xy
```
(where `xy` is substituted with the new target).

sanidhya00081:
Hey @cyrille-artho I have added a new test to specifically target the StaticCallToNonStaticTest.java. The final build was also successful. So, could you please check it and provide me with needful suggestions.

cyrille-artho (changes requested):
Please also add the new test target to the GitHub action, so the additional test is executed after the regular unit tests have passed.
Please remove `lib`, as the library dependencies are specified in `build.gradle` and downloaded automatically.

sanidhya00081:
Hey @cyrille-artho  I have made the changes assuring the successful build of new test and the old tests.

cyrille-artho:
Thanks! I can see that the new Gradle target `runModifiedTest` with the test gets started, but for some reason, it does not find the test. There is no JUnit output in that target (and that new test suite should fail until we have a fix). Can you please take a look?

sanidhya00081:
@cyrille-artho  Now my now test is targeting `StaticCallToNonStaticTest()`. And the output says 
```
StaticCallToNonStaticTest > testStaticCallToNonStatic() STANDARD_OUT
    Static method m() is called!
```
The build was also successful with successful running of other 1010 tests.

sanidhya00081:
I checked it with method `m()` in `D.java` as first non-static than static @cyrille-artho .

sanidhya00081:
Hey @cyrille-artho  could you please check the changes.

src/tests/gov/nasa/jpf/StaticCallToNonStaticTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class StaticCallToNonStaticTest extends TestJPF {
+
+    @Test
+    public void testStaticCallToNonStatic() {
+  D.m();
+}
+
+    // Cleanup method to delete the temporary file after tests
+    @AfterClass
```
cyrille-artho, line 19: Is this cleanup function needed? Your new test does not create a temp. file directly. If this happens due to some side effect somewhere, we can ignore it (and add that file to `.gitignore`).

src/tests/gov/nasa/jpf/StaticCallToNonStaticTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.File;
+
+
+
+
+public class StaticCallToNonStaticTest extends TestJPF {
+
+    @Test
+    public void testStaticCallToNonStatic() {
+  D.m();
```
cyrille-artho, line 15: This test should expect an exception: Try
`Assertions.assertThrows(java.lang.IncompatibleClassChangeError, D.m());`
This should make the test fail, so we can ensure that the fix that we need is tested against.
After the fix is in place, the test will succeed.

cyrille-artho (changes requested):
Thanks, I can see that the test is now executed.
Please make sure the test expects an exception of the right type, as it currently passes due to JPF not throwing the required exception.
Also, try to remove `settings.json` and the "cleanup" function unless the latter is really needed.

sanidhya00081:
Hey @cyrille-artho  please check my last commits. All the things work as intended. The test first fails and after the correction it passes. I have updated the clean-up thing as well as `settings.json`.

sanidhya00081:
@cyrille-artho  The failed build has occurred as to show that the test runs against incorrect function.

cyrille-artho:
Great job, we now have a test that reproduces the problem. Next, we want to fix the issue.
Most likely, the fix is to be made in `src/main/gov/nasa/jpf/jvm/bytecode/INVOKESTATIC.java`, in method `execute`.
This code handles method calls to static methods. It currently lacks a check to ensure that the invoked method is actually static.
The `callee` is a JPF-internal data structure of type `MethodInfo`. Its method `isStatic` should be an easy way to check whether the invoked method is indeed static when expected to be. In that case, JPF should throw an exception *at the level of the application under test* (and not within itself).
You can see that you do not want to throw an exception in the usual way, but luckily, the code in `execute` already has a similar example:
```
    if (callee == null) {
      return ti.createAndThrowException("java.lang.NoSuchMethodException", cname + '.' + mname);
    }
```
Helper method `createAndThrowException` can build the right type of object and return its descriptor in a way that the application will then see a newly constructed exception.
You can add a check on whether the `callee` instance is static and then throw an exception, very similar to the code above. Of course, the type should be "java.lang.IncompatibleClassChangeError".
We are just a few lines away from resolving this issue now!

cyrille-artho:
Yes, please try the fix I outlined; if it works, all old and new tests will pass.

cyrille-artho:
Your suggested change looks good. Where in the code did you put the check? What is the result?
You can update your PR with the latest change, which is the easiest way for me to see and test your exact change.

sanidhya00081:
@cyrille-artho  In my PR you could see that I added the check in` INVOKESTATIC.java` as suggested. But else everything works fine when I make `static` in `D.java`. But when I remove it I think compiler only is rejecting the method saying `static reference to non-static function` and `INVOKESTATIC.java` is not working in this case.

src/tests/gov/nasa/jpf/test/D.java
```diff
@@ -0,0 +1,9 @@
+// src/tests/gov/nasa/jpf/test/D.java
+package gov.nasa.jpf;
+
+public class D {
+    // Non-static method
+    public void m() {
```
cyrille-artho, line 6: This method should be static but then be compiled separately as a non-static method after an edit (which is made by `build.gradle`).

cyrille-artho (commented):
Method `D.m` needs to be static when the main source tree is compiled, otherwise, compilation fails.

sanidhya00081:
After making `D.java` `static` the build runs fine @cyrille-artho . But I am not able to understand that why `INVOKESTAITC.java` is not throwing exception. Or how the static check in `INVOKESTATIC.java` helps in our build process.

Issue #313, opened 2022-01-22
New Methods are missing in String Class


gaurangkudale:
Java 11 adds a few new methods to the String class: isBlank, lines, strip, stripLeading, stripTrailing, and repeat.
These methods are missing in this path `jpf-core/src/classes/java/lang/String.java` which is supported by java 11

cyrille-artho:
We tend to add methods to model classes when needed. Some of these functions are not so widely used. If someone provides a pull request that also has a unit test to cover the new method, we will of course gladly merge it.

gaurangkudale:
> We tend to add methods to model classes when needed. Some of these functions are not so widely used. If someone provides a pull request that also has a unit test to cover the new method, we will of course gladly merge it.

for this, we first need to add the methods in the model class and then unit test is that right?

cyrille-artho:
You would normally add the unit test first, to see that the unit test fails (and thus achieves something). The newly added method at step 2 would then fix the unit test.
The pull request then includes both the new test and new method (or class). You can commit your code in between steps without making a pull request.

AnkurRanpariya2005:
Please assign me

cyrille-artho:
I think there is no need, as this issue has not been active. You can make one PR for several missing methods or several PRs with one or a few methods each, whatever suits you best. Please ensure that in each PR, all missing methods are used by at least one test case that also checks the result of the method.

OmPradnya:
I went ahead and implemented the lines() method and also the corresponding unit tests, as it was the only missing String method in Java 11 in this context.

I have opened a PR for it here:

Please let me know if at all any changes are needed. Or any other gaps i should look into.

PR #512, opened 2025-01-29
feat: Add Java 11 String methods to JPF String class

gitkaarma:
This PR adds implementations for the new String methods introduced in Java 11 to the JPF String class. The following methods have been added:

- isBlank()
- strip()
- stripLeading()
- stripTrailing()
- repeat(int count)


gitkaarma:
hi @cyrille-artho, can you please review this PR related to the issue https://github.com/javapathfinder/jpf-core/issues/313

cyrille-artho:
Thank you!

PR #515, opened 2025-02-06
add unit test for jpf-core for supporting jdk17 new features

jerryyummy:
add some unit tests for jdk17 new features

### DIFF
--- a/src/tests/java17/BigIntegerTest.java
+++ b/src/tests/java17/BigIntegerTest.java
@@ -1,0 +1,16 @@
+package java17;
+
+import gov.nasa.jpf.util.test.TestJPF;
+import org.junit.Test;
+
+import java.math.BigInteger;
+
+public class BigIntegerTest extends TestJPF {
+    @Test
+    public void testBigIntegerMethods() {
+        if (verifyNoPropertyViolation()) {
+            BigInteger num = new BigInteger("97");
+            assertTrue(num.isProbablePrime(10));
+        }
+    }
+}
--- a/src/tests/java17/CollectionTest.java
+++ b/src/tests/java17/CollectionTest.java
@@ -1,0 +1,23 @@
+package java17;
+
+import gov.nasa.jpf.util.test.TestJPF;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CollectionTest extends TestJPF {
+    @Test
+    public void testCollectionsMethods() {
+        if (verifyNoPropertyViolation()) {
+            List<String> list = List.copyOf(Arrays.asList("A", "B", "C"));
+            assertEquals(3, list.size());
+
+            // `Collectors.toUnmodifiableList()`
+            List<String> immutableList = List.of("X", "Y", "Z")
+                    .stream().collect(Collectors.toUnmodifiableList());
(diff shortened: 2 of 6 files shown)


cyrille-artho:
Thank you for adding these tests. They can be useful to increase coverage.
However, most of the new tests currently fail. Please look into the output. Furthermore, the features you test are already present in Java 8, so it is best to make a pull request against `master`. We use Java 11 as default there, and all tests pass, so you can easily see if a new test causes the test suite to fail.
I will add a few more comments on details in GitHub's review feature.

```
java17.BigIntegerTest > testBigIntegerMethods FAILED
    java.lang.AssertionError at BigIntegerTest.java:11

java17.CollectionTest > testCollectionsMethods FAILED
    java.lang.AssertionError at CollectionTest.java:13

java17.FileTest > testFileMethods FAILED
    java.lang.AssertionError at FileTest.java:12

java17.HttpClientTest > testHttpClient FAILED
    java.lang.AssertionError at HttpClientTest.java:14
```
and
```
java17.StringTest > testStringMethods FAILED
    java.lang.AssertionError at StringTest.java:9
```

src/tests/java17/BigIntegerTest.java
```diff
@@ -0,0 +1,16 @@
+package java17;
```
cyrille-artho, line 1: Please put the new tests into the namespace of the regular tests, as they enhance tests on existing features rather than covering something that was introduced with Java 17.

src/tests/java17/StrongEncapsulationTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.Test;
+import java.lang.reflect.Method;
+
+public class StrongEncapsulationTest extends TestJPF {
+
+    @Test
+    public void testAccessPrivateMethods() {
+        try {
+            Method method = String.class.getDeclaredMethod("valueOf", Object.class);
+            method.setAccessible(true);
```
cyrille-artho, line 13: Method `valueOf` is public by default, so calling `setAccessible` does not change anything.
A `SecurityManager` setting may deny the usage of `setAccessible`, but we do not support that feature in JPF at the moment.

src/tests/java17/CollectionTest.java
```diff
@@ -0,0 +1,23 @@
+package java17;
```
cyrille-artho, line 1: Same here and for the other classes; please add them in the existing namespace, as we want to use `javaXX` only for features that specifically introduces in a given Java version.

src/tests/java17/StrongEncapsulationTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+            String result = (String) method.invoke(null, 42);
+            assertEquals("42", result);
+        } catch (Exception e) {
```
cyrille-artho, line 17: If the code works as expected, no exception should occur; therefore, the `try`/`catch` block can be removed (so the test fails with an exception if the code does not behave as expected). You only need to keep lines 12–16 from the method body.
Don't forget to add `if (verifyNoPropertyViolation()) {`, though; otherwise, the test runs on the host JVM and does not test JPF itself.

cyrille-artho (changes requested):
Thank you for the new tests.
Please see my comment to the PR and the comments here.

PR #513, opened 2025-01-30
Handle internal function signatures correctly for functions used by records in Java 17

eklaDFF:

PR #520, opened 2025-03-12
Record support

Mahmoud-Khawaja:
### Ongoing Work on Record Support

- Implemented `parseRecordAttribute()`.  
- Added support for printing record components in `ClassFilePrinter`.


Mahmoud-Khawaja:
Hello @cyrille-artho,

Hope you're doing well. I’ve added the foundations for records in Java 17 so the VM can recognize records and access their components. However, I’m currently stuck on implementing the implicit methods like toString(), hashCode(), and constructor.

Could you let me know if anything needs to be changed and guide me on the next steps? 

Thanks!

cyrille-artho:
FYI, support for internal methods depends on the correct function signature handling for bootstrap methods; see PR #513. That PR is work in progress so the feature does not work yet.

PR #521, opened 2025-03-16
Added check to accept version 61, throw error for higher

Mahmoud-Khawaja:


Mahmoud-Khawaja:
Hello @cyrille-artho,
i noticed that there is no explicit check for the major version. so if we tried to parse a class of any version jpf will accept it leading to potentail errors. if we checked the class file versions from here : [https://javaalmanac.io/bytecode/versions/]( )
we can find that java 17 works with class file version 61.0. so i added a check to prevent any kind of error later.

cyrille-artho:
Hi Mahmoud,
Thank you for this PR. Can you please add a test that shows that the new check is working? You might have to add a `.class` file that you compiled offline with Java 21, which then causes the version check to fail. The unit test should then expect a failure of the version check in order to pass.

Mahmoud-Khawaja:
im thinking of creating a `FileVersionTest.java` that has two tests on to test the supported and the other to test the unsupported version. then compile one time with java 21 and java 17 that should create two `.class` files
( fileVersionTest_java21.java,fileVersionTest_java17.java) if we if we disassembled `.class` for both one should have major version of 65 beacause we are compiling with java 21 and the other with major version of 61. so this what you mean?

cyrille-artho:
Yes, that is what I mean. It may be difficult to compile one file with two versions of `javac`, as the typical workflow of Gradle-supported builds is designed for one compilation per source file. For us, either way will work: dual compilation of one file or addition of a pre-compiled `.class` file to the repository as data.

cyrille-artho:
Together with a suitable test, try to make this PR against `master`, which is currently on Java 11.
This functionality is useful for that version, too, and we have 100 % passing tests on `master`, which makes it easier to review and merge new test cases and functionality.

Mahmoud-Khawaja:
Great! I'm currently working on this. I will commit the changes for you to review once I finish.

Mahmoud-Khawaja:
Hello @cyrille-artho ,
i've just implemented the unit tests. as you can see if we uncommented testUnsupportedVersionJava21() and tried to run the test with java 17 it will throw `ClassParseException` but it passes for testSupportedVersionJava17()

src/tests/gov/nasa/jpf/jvm/FileVersionTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        assertTrue("Java 17 class file should be parsed successfully", true);
+    }
+
+    /*
+    @Test(expected = ClassParseException.class)
+    public void testUnsupportedVersionJava21() throws IOException {
+        byte[] classData = loadClassFile(JAVA21_CLASS);
+
+        ClassFile classFile = new ClassFile(classData);
```
cyrille-artho, line 47: Please enable this test and ensure it throws an exception. The test looks good; what was the reason for commenting it out? Perhaps JPF loads the version of the class file that was compiled by the build process rather than the one in `src/tests`, which would explain that this test does not throw an exception and thus fails?
In that case, try to find a way to make your manually uploaded class file take precedence in the class loader.

cyrille-artho (changes requested):
Another item to address regarding the tests.

Mahmoud-Khawaja:
Hello @cyrille-artho,

thanks for response, you're correct about removing `TestClassJava17.class`. but for now I've been struggling with a problem for hours. i have updated the tests case and i think it should be working correctly but as you can see in `testUnsupportedVersionJava21()` this should throw `ClassParseException` since we are parsing a compiled file of class version 65 but im getting `AssertionError` after alot of debugging i tried to print the major version for classes using `getMajorVersion()` and for both tests it was 61 but the `testSupportedVersionJava17` passes which is okay in that case. 
but the other not throwing the expected error. even thoug i used ` javap -verbose` to check the byte code and its correct for both `.class` files. one has class version 61 and the other is 65.

after debugging again i tried to removed `TestClassJava17.java` and  `TestClassJava17.java` and kept the compiled files. it throwed an `IOException` which means we are not even loading the `.class` files. its a strange behaviour and im not sure what is it but when i got `TestClassJava17.java` and `TestClassJava21.java` back again and removed the compiled files. i got the same problem `testSupportedVersionJava17` passed and `testSupportedVersionJava21` throwed `AssertionError` 

so i think the problem is with the Class loading itself and im not sure why it can't read the compiled files 
I kept `TestClassJava17.class` to analyze the behavior for now, but I'm stuck with this problem.

can you please help me with this?

Mahmoud-Khawaja:
got it! I will investigate this, and maybe I can find the error

thanks!

Mahmoud-Khawaja:
Hello @cyrille-artho,
after debugging and searching for gradle standard behavior, i found out that `.class` files should be in a resources directory so i updated the `build.gradle` file to include `.class` files in the class path. and now it works correctly. 
let me know if anything needs to be modified. thanks a lot for your help!

Mahmoud-Khawaja:
yes, i forgot to add the .class now i think they will pass.

Mahmoud-Khawaja:
regarding the other failing tests i will investigate and see why they fail

PR #522, opened 2025-03-16
Fixes #355: Added Method dispatch test scenarios

### LINKED ISSUE #355: Add unit tests for method dispatch
We currently do not exhaustively test all possible configurations of method calls. It would be advantageous to have an overview of which tests are already there and which cases are covered:

1. Method call within the same class or to a superclass or inner class. (Three options)
2. Method call from/to instance/static method. (Four options)
3. Access modifiers. (Four options for caller and callee: 4^2).

Theoretically, 3 * 4 * 4^2 = 192 options are possible, but some are not allowed (a public superclass method must remain public).
A first step here is to model which combinations are possible and then write a table to see which cases are already tested. We want to test key cases of the remaining combinations, achieving at least pairwise coverage.

Mahmoud-Khawaja:


Mahmoud-Khawaja:
Implemented the first three test scenarios of. we can add some unit tests later 

| CallContext | CallerType | CalleeType | CallerAccess | CalleeAccess |
|-------------|------------|------------|--------------|--------------|
| same        | instance   | static     | default      | private      |
| inner       | instance   | instance   | default      | protected    |
| same        | instance   | instance   | private      | public       |
| inner       | static     | static     | default      | default      |
| same        | static     | static     | public       | public       |
| super       | instance   | instance   | default      | public       |
| super       | instance   | instance   | public       | default      |
| same        | static     | static     | private      | protected    |
| same        | static     | static     | private      | default      |
| inner       | instance   | instance   | protected    | private      |
| same        | instance   | instance   | private      | private      |
| inner       | instance   | instance   | public       | protected    |
| same        | static     | static     | protected    | protected    |
| super       | instance   | instance   | protected    | default      |
| inner       | instance   | instance   | protected    | public       |
| inner       | instance   | instance   | public       | private      |
| same        | static     | static     | protected    | private      |

cyrille-artho:
Thanks!

PR #523, opened 2025-03-16
Unit test enhancement

Mahmoud-Khawaja:
now testProxyCreationInCaseOfChoiceGenerator() is faster: 

in the original approach we were creating the thread then the proxy in the main thread. but now we are creating the proxy in the main thread then the thread itself. so this can reduce conncurrent operation and minimize the satate space. 

and when it comes to verification order the original approach verifying the the cross thread equality then the class format now i switched it as it can front-loads complex verification in single-threaded contex. 

the old test were taking up to 14 seconds now its about ~1 second

### DIFF
--- a/src/tests/gov/nasa/jpf/test/vm/reflection/ProxyTest.java
+++ b/src/tests/gov/nasa/jpf/test/vm/reflection/ProxyTest.java
@@ -134,13 +134,19 @@ public class ProxyTest extends TestJPF {
 
   @Test
   public void testProxyCreationInCaseOfChoiceGenerator() {
-    if (verifyNoPropertyViolation()){
+    if (verifyNoPropertyViolation()) {
+      MyHandler handler = new MyHandler(42);
+      Ifc ifc = (Ifc) Proxy.newProxyInstance(Ifc.class.getClassLoader(), new Class[] { Ifc.class }, handler);
+      String interfaceName = Ifc.class.getName();
+      String packageName = interfaceName.substring(0, interfaceName.lastIndexOf('.'));
+      String desiredProxyClsName = packageName + ".$Proxy$"
+              + Integer.toHexString(Ifc.class.getName().hashCode());
+      assertEquals(ifc.getClass().getName(), desiredProxyClsName);
+
+      // now test cross-thread behavior with minimal operations
+      // which mean we need fewer interleavings
       NewThread t = new NewThread();
       t.start();
-      MyHandler handler = new MyHandler(42);
-      Ifc ifc = (Ifc) Proxy.newProxyInstance(Ifc.class.getClassLoader(),
-                                             new Class[] { Ifc.class },
-                                             handler);
 
       try {
         t.join();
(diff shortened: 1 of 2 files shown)


Mahmoud-Khawaja:
Hello @cyrille-artho,
i don't why change_clock_duration_test fails here i haven't done any changes for this one. can you please help?

src/tests/gov/nasa/jpf/test/vm/reflection/ProxyTest.java
```diff
@@ -134,13 +134,19 @@ public void run() {
...
+      String packageName = interfaceName.substring(0, interfaceName.lastIndexOf('.'));
+      String desiredProxyClsName = packageName + ".$Proxy$"
+              + Integer.toHexString(Ifc.class.getName().hashCode());
+      assertEquals(ifc.getClass().getName(), desiredProxyClsName);
+
+      // now test cross-thread behavior with minimal operations
+      // which mean we need fewer interleavings
       NewThread t = new NewThread();
       t.start();
-      MyHandler handler = new MyHandler(42);
```
cyrille-artho, line 140: If you move all code of the current thread up to before the new thread is created, there is nothing left to interleave for JPF. It may be the case that this test cannot be optimized without losing the fact that it verifies thread interleavings here.

cyrille-artho (changes requested):
Good work; in one case, you have probably optimized too much. Try to see if you can keep some interleavings in `ProxyTest`; if the only way is to keep all interleavings, then we just revert the change I commented on (and keep the test as slow as it is now).

cyrille-artho:
Very few tests fail occasionally, but luckily, only rarely. I have restarted the CI build, and while waiting found that you probably removed the thread interleaving that the unit test in `ProxyTest` wanted to have.

Mahmoud-Khawaja:
i see, this one is hard to optimized as it will risk compromising the thread interleaving verification. However, i think the most optimization we can do is to pre compute the expected class name it can reduce the time but i don't think it will be a significant imrpovement. it can be something like ~3-4 seconds. i will try to do my best here to see if further significant improvements can be done 

Thanks a lot for your help!

cyrille-artho:
Sometimes the required complexity of a test is intrinsic (it's in the nature of the problem) and cannot be reduced further. In that case, that's fine, and at least we have reviewed the test now and know the complexity is not accidental (being caused by a poor approach to the problem).

syheliel:
> Hello @cyrille-artho, i don't why change_clock_duration_test fails here i haven't done any changes for this one. can you please help?

Hi, I find this case fails in my machine also 🤣

cyrille-artho:
Does that test fail consistently on your machine, or occasionally? In the CI setup on GitHub, it fails only very rarely so we can keep the test. If it fails consistently on some systems, it is worth investigating this. If we can't fix the problem, we might remove the test instead.

Mahmoud-Khawaja:
It works on my machine no problem with it.

syheliel:
I'm using ubuntu 24 under `vmware ESxi`. At first, the test will fail no matter how many times I try. But now it works again(trying to repeat running for 200 times)... I think it may related to some time sync problem for VM and currently no need to fix

PR #524, opened 2025-03-16
Java-17

sanidhya00081:


cyrille-artho:
Work on Java 17 is not complete yet, so we are not ready to make this the default version. Perhaps after this summer?

PR #527, opened 2025-03-21
enable parallel testing

syheliel:
This change reduces testing time from 5 min to 45 second


cyrille-artho:
I see only a very minor speed-up on the CI build and no significant change on my laptop (perhaps even a slowdown?); I will have to test this more on other machines.

syheliel:
I've do some research about this problem, there are 3 import parameters:
- max-workers: how many workers will be used, default value is cpu core number
- test.maxParallelForks: how many tests can be parallel executed in `Test` task, only work under `Test`
- test.forkEvery: whether to run every test under a sperate JVM.

`test.maxParallelForks` is bounded by `max-workers`. The default setup in github action is 4core, 16GB Memory, see [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). So the parallel test will not speed up a lot when considering the fork overhead... But it will help a lot in a server-level CPU with more cores.

syheliel:
Also the parallel level is showed when testing, like the folowing photo, 11 lines starting with `:test > Executing` indicates that there are 11 test runner
image

cyrille-artho:
Thanks. I have access to a machine with 128 cores, but it's currently down for maintenance. I hope to test your patch there next week. I would actually rather not cap the number of threads to 16 for that reason.
I think a problem is when using older machines (dual core processors). The base system load and overhead of managing threads outweighs the benefits there.
A solution could be to reduce the number of threads by one (from the maximum), using a minimum of one thread.

cyrille-artho:
Thank you for your patience, the larger machine is finally back online. I was able to confirm a big speed-up, but some tests are failing with the new settings:
```
gov.nasa.jpf.test.vm.basic.AssertTest > testSetDefaultAssertionStatus FAILED
    java.lang.AssertionError at AssertTest.java:128
...
gov.nasa.jpf.test.vm.basic.AssertTest > testClearAssertionStatus FAILED
    java.lang.AssertionError at AssertTest.java:156
```
Other tests in that package also seem to be flaky and may fail, due to `movePkg` in `LoadUtility`, the super class of `AssertTest`. That function moves files, which obviously wreaks havoc if multiple tests perform uncoordinated moves in parallel.
Is there a way to restrict parallel test executions in classes that do not allow it (by their test design)? Otherwise, we'd have to rewrite the tests to work without `movePkg`, which could be a major effort in itself.
The speed-up would be worth it, though, as it is nearly linear on a machine with many cores (even without removing the artificial limit of 16 cores).

cyrille-artho:
It seems that JUnit now has an annotation feature that makes it easy to exclude some tests from parallel test execution while keeping the rest parallel. The copied part below is AI-generated, but it looks plausible. If we add a `slow` tag to `AssertTest`, we might be able to run all the other tests in parallel and gain the desired speed-up without tests breaking.

```
// In your test class
import org.junit.jupiter.api.Tag;
import org.junit.jupiter.api.Test;

@Tag("slow")
class SlowTest {
    @Test
    void someSlowTest() {
        // ...
    }
}

// In your build.gradle file
test {
    useJUnitPlatform {
        excludeTags 'slow'
    }
    maxParallelForks = Runtime.runtime.availableProcessors() //Enable parallel execution for other tests
}

task slowTest(type: Test) {
    useJUnitPlatform {
        includeTags 'slow'
    }
    maxParallelForks = 1 // Ensure slow tests run sequentially
}
```

cyrille-artho (changes requested):
See my comment. Please try to exclude `AssertTest` from parallel testing, either with the suggestion I copied (if that works) or by another way of running tests twice.
An alternative way can be to first exclude `AssertTest` using a Gradle option and then again running Gradle to test only tests within `AssertTest`.

syheliel:
@cyrille-artho got it!

cyrille-artho:
Great! It seems your branch has conflicts with the main repo; please try to merge the latest changes from it into your work.
I recommend creating a feature branch (e.g., `parallel`) in your repo first so you have your current state saved there. It's also easier to make a PR from a feature branch.

build.gradle
```diff
@@ -270,38 +270,97 @@ tasks.register('buildJars') {
...
                                  "${result.successfulTestCount} passed",
                                  "${result.failedTestCount} failed",
                                  "${result.skippedTestCount} skipped"]
-
-            println "Summary: " + summaryFields.join(", ")
+            println "Summary (${name}): " + summaryFields.join(", ") // Use task name
         }
     }
+
+    include "**/*Test.class" // Include all potential tests first
```
cyrille-artho, line 356: Limit this to URLClassLoaderTest and AssertTest for the single-threaded case, so the test method lookup of JUnit's test runner will be fast.

cyrille-artho (changes requested):
Great job! Please check the merge conflicts and try to limit the number of classes in the single-threaded test set to a minimum. It turns out that going through all the test classes to check them for the right test methods is very slow (due to reflection being slow), so eliminating 99 % of these lookups should make it possible to improve performance about four-fold.

cyrille-artho:
Works now and is also fast, thank you very much!

PR #528, opened 2025-03-21
Class version check 

Mahmoud-Khawaja:


Mahmoud-Khawaja:
Hi @cyrille-artho,
The test passes on the master branch. Let me know if anything else needs to be done.

cyrille-artho:
Thank you very much!

PR #529, opened 2025-03-22
Completed TODO- MatchResult()

jayshah1819:
Completed the TODO methods in the Java Matcher class, implementing toMatchResult().

### DIFF
--- a/src/classes/modules/java.base/java/util/regex/Matcher.java
+++ b/src/classes/modules/java.base/java/util/regex/Matcher.java
@@ -159,7 +159,7 @@ public class Matcher {
   }
 
 
-  // TODO public native MatchResult toMatchResult();
+  // TODO public native MatchResult toMatchResult();-Done;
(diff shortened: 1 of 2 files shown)


syheliel:
Thanks for your contribution :) would you like to add some unit test for it?

jayshah1819 (commented):
Added Test Cases

syheliel:
@jayshah1819 Thanks! that's close, but the file inside `java.lang` is basically a "hook" to intercept the java's default standard library. You may want to make the test class extends `TestJPF` and wrap the test function with ` if(verifyNoPropertyViolation()){...}`. You can check this test for example: https://github.com/javapathfinder/jpf-core/blob/master/src/tests/java11/StringConcatenationTest.java

jayshah1819:
Thanks! I’ve updated the test class to extend TestJPF and wrapped the test function with if(verifyNoPropertyViolation()). Please let me know if I am missing anything!

cyrille-artho:
Thank you!

PR #530, opened 2025-03-23
Adding support for internals

Mahmoud-Khawaja:


Mahmoud-Khawaja:
Hello @cyrille-artho,
i noticed the issue here in this pr #513 was descriptor misuse as it was passed directly to `setBootstrapMethodInfo` as both the method descriptor and the bootstrap descriptor. also for the component index uses uses `cpArgs[0]` for `bmArg`
as constant pool index for record components might not always be the first argument

however i corrected this in the new implementaion and now it works well 
as you can see i descriptor to standard signatures to ensures `methodName` and `methodDescriptor` align with Java’s record contract.

i also implemeted the logic for .equals() and .hashCode() and it passed the unit test. 
the failing tests are realted to nested recorod equality and ToString for nested records and im currently working on this as i haven't implemented the logic for `.toString()` yet.

Mahmoud-Khawaja:
I've just implemented toString function. now only 6 tests fail with nested record equalty and generic record. i think i have to figure out how to handle the equlaity for generic records and nested record.

cyrille-artho:
Thank you for your great work. It took a while to review the PR, as it's fairly large. Please check or comment if there is a more direct way to deduce the right method in `objectMethodsBootstrap`.

Mahmoud-Khawaja:
Hello Cyrille,
thanks for the feedback. i left the debugging statments for now till i complemete the support for all functionalities. once i finish i will remove them. 
also i think we had 3 failing tests before adding the tests for generics and nested records. but know i will work on them as soon as i complete my midterms.
thanks!

Mahmoud-Khawaja:
Hi @cyrille-artho ,

after a lot of investigation regarding the failing tests. i noticed that my current implementation producing different hashcodes for the same array. like `new int[]{1, 2, 3}` and `new int[]{1, 2, 3}` (different objects) have different identity hash codes, even though their contents are equal. and the major issue that i wasn't checking the component type
`eg  "Z" - > boolean, "I" -> int, etc. so now im thinking to refactor the current implementation to care about the deep copy and take the primitive types in consideration. i look at `ElementInfo.java` and i found that we can get each primitive element like "getBooleanElement(),getIntElement(),etc". but the problem is that i think alot of if and else statements in my implementation so i was also thinking about modifing ElementInfo to find a way to map all primitive types and get it directly without any need of unnecessary checks or do you think we can leave it as it is for now. i just need to know if im on the right tack doing this 

thanks

cyrille-artho:
Hi,
If I read your post correctly, you are looking at two things:
1. `identityHashcode`. This is essentially a unique identifier that is set when an object is allocated. It should be different for two different objects with the same content! Contrarily, `hashCode` should be equal if the content is equal. In general, you can always try a small test on the regular JVM to see what the behavior should be.
2. Refactoring or providing a convenience method for primitive types is worth it if more than a few cases are needed. If we only need `I` and `Z`, a few `if` statements are fine. The main thing when refactoring is to follow the recommended workflow so you fix bugs or add new features together with a refactoring.

Refactoring workflow:
1. Finish any work in progress on the branch you work on. On `master`, this means 100 % passing tests; on WIP branches, take note of the (few) tests that currently fail, to ensure that this does not change after the refactoring.
2. Refactoring the code, and run the tests again to ensure their outcome does not change.

The workflow is simple, and the important part is not to mix refactorings with other things: If you find a way to fix a failing test, finish the refactoring first, make a PR, and then fix the test in a separate step. (If you find a way to fix a test before doing the refactoring, you can of course also do that first and make a PR with the fix before attempting the refactoring.)

Mahmoud-Khawaja:
Hi @cyrille-artho,

I’ve refactored my `INVOKEDYNAMIC` implementation, and now `testRecordDirectFieldAccess` (and 2 other tests) are failing. I’m debugging them and could use your insight.

For `testRecordDirectFieldAccess`, it tries to modify a record’s `final` field (`x` in `Point`) via reflection (`Field.set`) after `setAccessible(true)`. In the JVM, this throws `IllegalAccessException` because record fields are `final`, but in JPF, the modification succeeds, causing the test to fail. It seems JPF doesn’t enforce `final` field immutability at runtime, though it should mimic JVM behavior for records.

I’m wondering: does JPF intentionally allow direct field access (ignoring `final` or `private`) for debugging or simplification in some cases? If so, this test might not align with that design and could be skipped for now. Otherwise, I plan to add reflection support to enforce `final` fields by throwing `IllegalAccessException` in `Field.set`.

What do you think should I keep this test and fix reflection, or is it not a priority?

Thanks!

Mahmoud-Khawaja:
i will also try to find elegant way to other than `if, else` statment in `deepEquals()`. i know the code might not be easy to read but it works for now

Mahmoud-Khawaja:
i was running through a silly bug as i was trying to map bootstrap method’s descriptor to a record method name `(toString, hashCode, or equals)` . the `objectMethodsBootstrap` mistakenly used the bootstrap method’s descriptor for example this `Ljava17/records/RecordEqualsTest$ArrayHolder;)Ljava/lang/String;` instead of the `invokedynamic`  instruction’s method name. so this is we had a trouble loading `ArrayHolder` however i corrected the implementation as i already provided the method name in invokedynamic

Mahmoud-Khawaja:
now there's nothing left except `testRecordDirectFieldAccess`. i don't think there's any reason for JPF to allow access to a final field but when I asked I was just curious why it hadn't been implemented earlier. i will try to add reflection support later. but for now please review my PR and let me know if there are any remaining changes needed.

Mahmoud-Khawaja:
Hello @cyrille-artho ,
please checkout the changes and tell me your insights about it. i guess now i need to work on the dynamic bootstrap instruction in invkedyanmic as it will be crucial for dynamic code generation so insetead of just interpreting bootstrap methods this will properly resolves and creates actual call site. i will need sometime work on this. also for now there is no failing tests other than the one related with the direct field access after i solve it and finalize this pr i will open a new pr to work on start working on `executeDynamicBootstrap`.I'll close the last PR I made yesterday to avoid any overlap after merging, since I'm using the same functionalities here

cyrille-artho:
Hi,
Good work, I think we can merge this PR so the next PR can work on direct field access in isolation, making it easier to review and manage.

cyrille-artho (approved):

PR #531, opened 2025-03-24
add missing dep

syheliel:
`compileTestJava` should reply on `compileModulesJava`. The vscode will throw an error when running a single test without this change


cyrille-artho:
Thank you!

PR #532, opened 2025-03-27
[WIP] add support for `lookup.findStatic`

syheliel:
`lookup.findStatic` is a feature in Java that allows developers to dynamically find and invoke static methods using a MethodHandle. This capability is particularly useful in scenarios requiring dynamic method invocation, lambda expression implementation, and bytecode manipulation tools.

Here is the first example I meant to analyze(I'll add into test when code is done)
```
import java.lang.invoke.MethodHandle;
import java.lang.invoke.MethodHandles;
import java.lang.invoke.MethodType;

public class MethodHandleExample {
    public static int addTen(int x) {
        return x + 10;
}
    public static void main(String[] args) throws Throwable {

        MethodHandles.Lookup lookup = MethodHandles.lookup();
        MethodType mt = MethodType.methodType(int.class, int.class);
        MethodHandle addTen = lookup.findStatic(
                MethodHandleExample.class,
                "addTen",
                mt
      );
        System.out.println((int)addTen.invokeExact(10));
    }

}
```

I'm currently diving into the codebase, so any guidance would be immensely appreciated. Here are some problems I'm facing:
1. In first `TODO`, how to get the instance of `MethodType`? The `MethodType.toMethodDescriptorString()` Will be further use to determine which eaxct method to use.
2. In second `TODO`: What value should I return? Should I create a `MethodHandle` object on the stack?

### DIFF
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_invoke_MethodHandles.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_invoke_MethodHandles.java
@@ -17,9 +17,56 @@
  */
 package gov.nasa.jpf.vm;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+
 import gov.nasa.jpf.annotation.MJI;
 
 public class JPF_java_lang_invoke_MethodHandles extends NativePeer {
+  static Class<?> getClass(MJIEnv env, int classRef) throws ClassNotFoundException {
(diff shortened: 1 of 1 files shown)


cyrille-artho:
Thank you for your work on the native peer functionality. This requires a good understanding of JPF's internals, so it can take a while (and some trial and error) to figure out how to access existing data structures or how to return results.
I'm currently traveling and have limited time to study PRs.
Can you please give more context? You mention two TODO items, but your PR has only one. In your existing TODO item, is the issue how to ensure that the function in question is private? You would specify that when writing the model class that declares the native method, as the native peer itself is only visible within JPF in the first place.

PR #539, opened 2025-04-30
enhanced bootstrap method handling

Mahmoud-Khawaja:


Mahmoud-Khawaja:
@cyrille-artho,
hello Cyrille, 
hope you're well,
i noticed that in the old implementation we were assuming the count of the constant pool agrs however i did some changes that can be better than the old implementation. i also added the dynamic type support to handle generic bootsrap methods and also generating  call site code for bootstrap methods while keeping infomation about the original constant pool args

PR #543, opened 2025-06-03
Enabled ParallelTesting

Mahmoud-Khawaja:


Mahmoud-Khawaja:
Hi Cyrille, 
I was going to discuss this with you in tomorrows meeting. as far as i now this annotation is for tests that is hard to be parallelized. but when i tried to run the tests on my machine nothing failed. so can you please tell me what is the importance and why we have to include it? 
thanks!

Mahmoud-Khawaja:
Thanks you so much. i've just inclueded the annotation

cyrille-artho:
Another thing that is necessary (at least with some setups) is to set the source file encoding correctly. With OpenJDK on an ARM Mac, the line `options.encoding = 'UTF-8'` by itself does not seem to be sufficient.
I currently have four tests failing there, three (likely) due to the encoding issue, and one due to parallel tests interfering with each other.

Mahmoud-Khawaja:
i think this BasicRecordTest.testRecordDirectFieldAccess fails because i made a branch starting from the pr "Adding support for internals" which didn't include the fix for the direct field access

Mahmoud-Khawaja:
i will also try to look at the issue for the source encoding

Mahmoud-Khawaja:
we may explicitly configure the encoding for each build phase and try to make some debugging statements to verify if its working

Mahmoud-Khawaja:
i tried to enforce utf-8 across the jvm buidling phases. so i added this 


```
import java.nio.charset.Charset

tasks.withType(JavaCompile) {
    options.encoding = 'UTF-8'
    options.compilerArgs += ['-encoding', 'UTF-8']
}

tasks.withType(Test) {
    systemProperty 'file.encoding', 'UTF-8'
    jvmArgs '-Dfile.encoding=UTF-8'
}

tasks.withType(JavaExec) {
    systemProperty 'file.encoding', 'UTF-8'
}

task validateBuildEncoding {
    doLast {
        println "JVM Encoding: ${System.getProperty('file.encoding')}"
        println "Default Charset: ${Charset.defaultCharset().name()}"
        println "Native Encoding: ${System.getProperty('native.encoding')}"
    }
}


```
When I ran validateBuildEncoding, I got the following output:

```
JVM Encoding: UTF-8
Default Charset: UTF-8
Native Encoding: UTF-8 
```


so i think it should be working this way. what do you think ?

cyrille-artho:
Unfortunately, I still get four failing tests. Please try again starting from branch `java-17`, which should work.

Perhaps:
```
git diff f2cec97a8fc442172e0bcc66a1e511007c15a8df > /tmp/parallel.patch
git checkout java-17 # back to java-17
git branch parallel # new branch based on java-17
git checkout parallel
patch -p1 < /tmp/parallel.patch # apply changes
```

cyrille-artho:
Thanks!

PR #544, opened 2025-06-03
refactored sealded classes tests

Mahmoud-Khawaja:


src/tests/java17/SealedClassesTest.java
```diff
@@ -0,0 +1,346 @@
...
+
+
+    // Basic Sealed Interface
+    sealed interface Shape permits Circle, Rectangle {
+        int value();
+    }
+    final class Circle implements Shape {
+        private final int radius;
+        Circle(int radius) { this.radius = radius; }
+        @Override public int value() { return radius; }
```
cyrille-artho, line 215: Maybe change to area and return the area here, too?
In that case, the radius should be a `float` parameter, and the check should be approximate (the result may not match exactly).

src/tests/java17/SealedClassesTest.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+    final class Rectangle implements Shape {
+        private final int width, height;
+        Rectangle(int width, int height) { this.width = width; this.height = height; }
+        @Override public int value() { return width * height; }
```
cyrille-artho, line 220: Here: area

cyrille-artho (changes requested):
Please rewrite the tests a bit so they do not match existing tutorials exactly.

PR #545, opened 2025-06-05
Sealed classes full

Mahmoud-Khawaja:


Mahmoud-Khawaja:
I’ll try to work on a test that throws an exception. It might be challenging, but I’ll do my best to figure it out.

src/main/gov/nasa/jpf/jvm/ClassFilePrinter.java
```diff
@@ -224,6 +224,22 @@ public void setRecordComponentAttribute(ClassFile cf, Object tag, int componentI
     pw.println();
   }
 
+  @Override
```
cyrille-artho, line 227: Please try to isolate this chunk and revert all the other whitespace changes, in case we have to merge this patch later on other branches like `java-11`.

cyrille-artho (approved):

PR #554, opened 2025-07-05
added direct execution for recrods

Mahmoud-Khawaja:


src/main/gov/nasa/jpf/jvm/bytecode/INVOKEDYNAMIC.java
```diff
@@ -158,25 +161,12 @@ private CallSite createCallSite(ThreadInfo ti, ClassInfo enclosingClass,
...
                                       BootstrapMethodInfo bmi) throws Throwable {
+    // call site generation works only in the context of string concat
     MethodHandle target = callSite.getTarget();
     MethodType type = target.type();
-
     debugLog("Executing CallSite type: " + bmi.getBmType());
     debugLog("Target type: " + type);
-
-    switch (bmi.getBmType()) {
-      case STRING_CONCATENATION:
```
cyrille-artho, line 168: I think it's good to keep this `switch` block even if the methods that are called from it are only stubs.
If you remove them entirely, won't other cases be treated as string concatenations and then fail?
If it's safe to remove this block, please remove the stubs, too (as with `executeDynamic`). Please comment either way why you can remove this block (in the discussion of the PR).

cyrille-artho (changes requested):
Please clarify if the `switch` block can be safely removed.

src/main/gov/nasa/jpf/jvm/bytecode/INVOKEDYNAMIC.java
cyrille-artho: Thanks for clarifying. I now had time to look at the code more and saw that the other `execute...` methods are still used elsewhere, so we don't get dead code with your change. I'll merge it.

cyrille-artho (approved):

PR #559, opened 2025-08-22
containerized jpf

Mahmoud-Khawaja:


cyrille-artho:
Thanks. Please add instructions on how to use/test the image. 

Also, `WORKDIR` is `/opt/jpf/jpf-core` but we have `/home/jpf` below:
```
      - ~/.gradle:/home/jpf/.gradle
```

cyrille-artho (approved):

eklaDFF:
`openjdk:11-jdk-slim` does not work now. Can be changed with `eclipse-temurin:11-jdk-jammy`.

cyrille-artho:
Please make a pull request so we can trigger the CI build and see if this works as expected.
It is a good idea to migrate away from any `-latest` images and versions, as these are not stable and can change without notice. Using a fixed version avoids this problem.

eklaDFF:
#567

PR #568, opened 2025-11-19
Optimize testRecursiveJoinThreadGroup performance

### LINKED ISSUE #418: Unit test case performance

HritikRaj2:
- Removed innermost thread t11 to reduce nesting from 3 to 2 levels
- Reduced thread count from 4 to 3
- Updated counter indices from 0-3 to 0-2
- Performance improvement: 5,853ms -> 3,962ms (32.3% faster)
- Test still passes and maintains full test coverage

Fixes #418

### DIFF
--- a/src/tests/gov/nasa/jpf/test/vm/threads/JoinTest.java
+++ b/src/tests/gov/nasa/jpf/test/vm/threads/JoinTest.java
@@ -400,7 +400,6 @@ public class JoinTest extends TestJPF {
       Verify.resetCounter(0);
       Verify.resetCounter(1);
       Verify.resetCounter(2);
-      Verify.resetCounter(3);
     }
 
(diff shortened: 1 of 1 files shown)


HritikRaj2:
Hi @cyrille-artho,

I'd like to continue the test optimization work from #418. I successfully optimized 2 tests:
- testRecursiveJoinThreadGroup: -32%
- testStackWalkerInCaseOfChoiceGenerator: -29%

May I continue optimizing more tests from the original slowest-50 list without opening new issues? I'll submit each as a focused PR with before/after metrics.

Thanks!

PR #576, opened 2025-11-29
Java 21 Migration

### LINKED ISSUE #575: Java 21 test fails
The current codebase compiles on the current LTS release Java 21 but there are 791 failing test cases
1. Gradle 8.4 is incompatible with Java 21.
2. JPF Core relies on `jdk.internal.*` and `SharedSecrets` accessors that were removed or refactored in Java 21.
3. Introduction of virtual threads and Thread Container

Image

I am testing on the Java-17 branch as following: https://gist.github.com/Sameersawant77/5649e23e4d6ddc599c64c512af63f5d2

Sameersawant77:
Fixes #575 

Refactoring and Migrating the codebase to support Java 21
I have reduced test failures from 791 down to 9. I am currently cleaning up the remaining edge cases and will open a PR shortly.

Changes:
Gradle Upgrade
Native Peers
Internal Accessors
Virtual Threads and Threads Container Handling

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM eclipse-temurin:21-jdk-jammy
+
+RUN apt-get update && apt-get install -y \
+    git \
+    unzip \
+    bash \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /home/jpf-core
+
+WORKDIR /home/jpf-core
+
+RUN git config --global --add safe.directory /home/jpf-core
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,10 @@
+services:
+  jpf-dev:
+    image: jpf-core
+    build:
+      context: .
+    working_dir: /home/jpf-core
+    volumes:
+      - .:/home/jpf-core
+      - ~/.gradle:/root/.gradle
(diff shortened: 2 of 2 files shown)


src/classes/modules/java.base/jdk/internal/access/SharedSecrets.java
```diff
@@ -17,10 +17,9 @@
  */
 package jdk.internal.access;
 import jdk.internal.misc.Unsafe;
-import java.io.File;
-import java.io.FileDescriptor;
-import java.io.ObjectInputFilter;
-import java.io.ObjectInputStream;
+
+import java.io.*;
```
cyrille-artho: Using wildcards is convenient but obscures which classes we actually import; please keep the explicit list of classes as before this change.

cyrille-artho (changes requested):
Great work!
Looks good overall; I have only a few minor comments.
However, please make your PR against branch `java-21`. I have created it remotely now.
We are still using `java-17` to support Java 17, and we have all tests passing there now.

cyrille-artho:
Looks good; please make a new PR with the minor changes I've requested to branch `java-21`. This way, we can keep support for Java 17 (hopefully soon with extensions, too).

PR #577, opened 2025-11-29
Fix VM.getSavedProperty('os.name') not supported

HritikRaj2:
- Add 'os.name' to system properties list in JPF_java_lang_System.java
- Implement jdk.internal.misc.VM class with native methods
- Add VM peer with saved property management in JPF_jdk_internal_misc_VM.java
- Initialize VM saved properties from system properties during VM init
- Add compilation exports for examples in source-sets.gradle
- Include test case VMPropertyTest.java to verify functionality

Resolves GitHub issue #495

### DIFF
--- a/src/classes/modules/java.base/java/lang/System.java
+++ b/src/classes/modules/java.base/java/lang/System.java
@@ -28,6 +28,7 @@ import jdk.internal.misc.SharedSecrets;
 import jdk.internal.reflect.ConstantPool;
 import sun.nio.ch.Interruptible;
 import sun.reflect.annotation.AnnotationType;
+import jdk.internal.misc.VM;
 
 
 public class System {
@@ -66,7 +67,7 @@ public class System {
     // would have prefered not to learn about. It's a mess WRT Java 1.5 / 6 compatibility
     // <2do> - most if this isn't supported yet
     SharedSecrets.setJavaLangAccess( createJavaLangAccess());
-
+    
     // <2do> this is an approximation that isn't particularly safe since we don't
     // initialize sun.misc.VM
     //sun.misc.VM.booted();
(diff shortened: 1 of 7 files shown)


src/classes/modules/java.base/java/lang/System.java
```diff
@@ -66,7 +67,7 @@ public class System {
     // would have prefered not to learn about. It's a mess WRT Java 1.5 / 6 compatibility
     // <2do> - most if this isn't supported yet
     SharedSecrets.setJavaLangAccess( createJavaLangAccess());
-
+
```
cyrille-artho: Revert this change (whitespace) so we can merge this change more easily with other branches in the future.

src/classes/modules/java.base/jdk/internal/misc/VM.java
```diff
@@ -0,0 +1,16 @@
...
+    private static native void initialize();
+    
+    public static native void saveAndRemoveProperties(Properties props);
+    
+    public static native String getSavedProperty(String key);
+    
+    static {
+        initialize();
+    }
+}
```
cyrille-artho: Newline.

src/peers/gov/nasa/jpf/vm/JPF_jdk_internal_misc_VM.java
```diff
@@ -1,41 +1,50 @@
...
-    // nothing here yet
-  }
+    @MJI
+    public int getSavedProperty__Ljava_lang_String_2__Ljava_lang_String_2(MJIEnv env, int clsObjRef, int keyRef) {
+        String key = env.getStringObject(keyRef);
+        String value = savedProps.getProperty(key);
+        return (value != null) ? env.newString(value) : MJIEnv.NULL;
+    }
 
-  @MJI
```
cyrille-artho: Please revert this change (which only reformats code) as to facilitate merging code with other branches.

cyrille-artho (changes requested):
Thank you for your effort. Your changes look good, aside minor formatting issues.
Unfortunately, several unit tests now fail. Please look into that.
Perhaps we can include a subset of your changes so the unit tests still pass and work on other features later.

PR #582, opened 2025-12-09
Use ZonedDateTime instead of LocalTime to handle midnight wrapping

Sameersawant77:
### Description
This PR fixes a flaky unit test (`change_clock_duration_test`) that fails consistently during the one hour before midnight.

### The Bug
The test previously used `LocalTime` to calculate the difference between two clocks. 
* **Scenario:** When the time is `23:00` (11 PM).
* **Logic:** `est` is `23:00`. The offset clock `edt` is `23:00` + 1 hour = `00:00`.
* **Failure:** `LocalTime` discards date information. It calculates the difference between `23:00` and `00:00` (on the same abstract day) as `+23 hours` instead of `-1 hour`. This caused the assertion `assertThat(timeDifference, is(-1L))` to fail.

### The Fix
Replaced `LocalTime` with `ZonedDateTime`. 
`ZonedDateTime` preserves the date information, so when the time wraps past midnight to the next day, the `until()` calculation correctly identifies the time difference as `-1 hour`.

### Verification
* Verified locally by mocking `Clock.fixed()` at `23:00` to reproduce the failure.
* Confirmed that `ZonedDateTime` resolves the assertion correctly at the midnight boundary.


cyrille-artho:
Thank you!

PR #583, opened 2025-12-09
Fix #341: Add model classes for java.util.logging

### LINKED ISSUE #341: Challenges running java.util.logging.Logger.getLogger in JPF
I have this super minimal reproduction:

```
public class Input {
    System.out.println("-------- In main!");

    java.util.logging.Logger.getLogger("blah");
  }
}
```

I'm running it against the `java-10-gradle` branch (a8160543fecaee799587c955c429aa7f068a410a) and the first thing I get when calling that is:

```
java.lang.NoSuchMethodException: Calling java.lang.Thread.<init>(Ljava/lang/ThreadGroup;Ljava/lang/Runnable;Ljava/lang/String;JZ)V
	at java.util.logging.LogManager$Cleaner.<init>(LogManager.java:261)
	at java.util.logging.LogManager.<init>(LogManager.java:301)
	at java.util.logging.LogManager.<init>(LogManager.java:294)
	at java.util.logging.LogManager$1.run(LogManager.java:248)
	at java.util.logging.LogManager$1.run(LogManager.java:223)
	at java.security.AccessController.doPrivileged(AccessController.java:34)
	at java.util.logging.LogManager.<clinit>(LogManager.java:223)
	at java.util.logging.Logger.demandLogger(Logger.java:648)
	at java.util.logging.Logger.getLogger(Logger.java:717)
	at java.util.logging.Logger.getLogger(Logger.java:701)
	at Input.main(Input.java:5)
```

So I added an extra method with an (ignored) boolean to here: https://github.com/javapathfinder/jpf-core/blob/a8160543fecaee799587c955c429aa7f068a410a/src/classes/modules/java.base/java/lang/Thread.java#L118-L121

I then ran into a similar issue from `java.security.AccessController` and added another dummy overload there.

That got me past the initial issue, but now I'm stuck here:

```
java.lang.UnsatisfiedLinkError: cannot find native jdk.internal.misc.VM.initializeFromArchive
	at jdk.internal.misc.VM.initializeFromArchive(gov.nasa.jpf.vm.JPF_jdk_internal_misc_VM)
	at java.util.ImmutableCollections$SetN.<clinit>(ImmutableCollections.java:583)
	at java.util.Set.of(Set.java:502)
	at java.nio.file.spi.FileSystemProvider.<clinit>(FileSystemProvider.java:428)
	at sun.nio.fs.DefaultFileSystemProvider.<clinit>(DefaultFileSystemProvider.java:35)
	at java.nio.file.FileSystems.getDefault(FileSystems.java:185)
	at java.nio.file.Path.of(Path.java:147)
	at java.nio.file.Paths.get(Paths.java:69)
	at java.util.logging.LogManager.getConfigurationFileName(LogManager.java:1389)
	at java.util.logging.LogManager.readConfiguration(LogManager.java:1375)
	at java.util.logging.LogManager.readPrimordialConfiguration(LogManager.java:445)
	at java.util.logging.LogManager$2.run(LogManager.java:394)
	at java.security.AccessController.doPrivileged(AccessController.java:39)
	at java.util.logging.LogManager.ensureLogManagerInitialized(LogManager.java:382)
	at java.util.logging.LogManager.getLogManager(LogManager.java:430)
	at java.util.logging.Logger.demandLogger(Logger.java:648)
	at java.util.logging.Logger.getLogger(Logger.java:717)
	at java.util.logging.Logger.getLogger(Logger.java:701)
	at Input.main(Input.java52)
```

Which doesn't seem to be as easy to fix as the two above. Anyone have hints for how to get past this?

HritikRaj2:
### Problem
JPF crashes when code calls `Logger.getLogger()` because the real JDK implementation uses:
- Native methods (`jdk.internal.misc.VM.initializeFromArchive`)
- Unsupported Thread constructors
- Complex security and filesystem APIs

### Solution
Created model classes for `java.util.logging.Logger`, `LogManager`, and `Level` that provide simplified, JPF-compatible implementations.

### Changes
- `src/classes/modules/java.logging/java/util/logging/Logger.java` - Model class for Logger
- `src/classes/modules/java.logging/java/util/logging/LogManager.java` - Model class for LogManager
- `src/classes/modules/java.logging/java/util/logging/Level.java` - Model class for Level
- `src/tests/gov/nasa/jpf/test/java/util/LoggerTest.java` - Test case

### Testing
All existing tests pass + new test confirms `Logger.getLogger()` works without crashes.


src/classes/modules/java.logging/java/util/logging/FileHandler.java
```diff
@@ -6,13 +6,13 @@
  * The Java Pathfinder core (jpf-core) platform is licensed under the
  * Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
- * 
- *        http://www.apache.org/licenses/LICENSE-2.0.
```
cyrille-artho: Please avoid changes in whitespace, as we subsequently want to merge changes to `master` to other branches (to support newer Java versions). Therefore, we want to keep only essential changes.

src/tests/gov/nasa/jpf/test/java/util/LoggerTest.java
```diff
@@ -0,0 +1,18 @@
...
+import gov.nasa.jpf.util.test.TestJPF;
+import org.junit.Test;
+import java.util.logging.Logger;
+
+public class LoggerTest extends TestJPF {
+
+    @Test
+    public void testGetLogger() {
+        if (verifyNoPropertyViolation()) {
+            System.out.println("--- calling getLogger");
```
cyrille-artho: remove println in unit tests

src/tests/gov/nasa/jpf/test/java/util/LoggerTest.java
```diff
@@ -0,0 +1,18 @@
...
+            Logger log = Logger.getLogger("testLogger");
+            assertNotNull(log);
+            System.out.println("--- success: " + log.getName());
```
cyrille-artho: Change this statement to an assertion (e.g., `assertEquals(log.getName(), "testLogger"`) and add a few assertions to check other log settings.

src/tests/gov/nasa/jpf/test/java/util/LoggerTest.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.logging.Logger;
+
+public class LoggerTest extends TestJPF {
+
+    @Test
+    public void testGetLogger() {
+        if (verifyNoPropertyViolation()) {
+            System.out.println("--- calling getLogger");
+            Logger log = Logger.getLogger("testLogger");
+            assertNotNull(log);
```
cyrille-artho: You can remove this line, as `log` is used below, so the test would fail if `log` were `null`.

cyrille-artho (changes requested):
Thank you for your contribution.
Please remove changes in whitespace in `FileHandler`, as there is no code change there.
Also, please try to add a minimal `Handler` for the log output so we can "write" the log to an array instead of to the terminal. This avoids cluttering the output during testing and lets us compare the output to the expected result.

src/tests/gov/nasa/jpf/test/java/util/LoggerTest.java
HritikRaj2: Thank you for the detailed review!

I will address all the points:
1. Reverting the whitespace changes in FileHandler.java.
2. Removing the System.out.println calls in the test.
3. Implementing a minimal `Handler` and `LogRecord` model so we can redirect output to a stream.
4. Updating the test to verify log content using assertions instead of visual inspection.

Working on the updates now!

PR #607, opened 2026-03-05
fix: remove duplicate isSealed and permittedSubclassNames fields from JVMClassInfo

### LINKED ISSUE #605: isSealed and permittedSubclassNames field shadowing makes sealed class enforcement non-functional
I've noticed something in the old implementation for sealed classes #545. even though the tests are passing but it seems that its non funtioncal. `JVMClassInfo` redeclares two fields that already exist in parent class `ClassInfo`:

```java
protected boolean isSealed;
protected String[] permittedSubclassNames;
```

the Initializer sets JVMClassInfo's copies, but `isSealed()` and `isPermittedSubclass()` in `ClassInfo` read the parent's copies which are always false and `EMPTY_STRING_ARRAY`. Sealed class enforcement in `ClassLoaderInfo` is therefore completely dead code. 

The fix is really straight forward 

remove the isSealed and permittedSubclassNames field declarations from JVMClassInfo.  They're already declared as protected in the parent ClassInfo. so the initializer can set them without redeclaring. Thats why this was preventing `isSealed()` and `isPermittedSubclass()` from ever seeing the values We're setting

Himaswin:
Fixes #605

Removed duplicate `isSealed` and `permittedSubclassNames` field declarations 
from JVMClassInfo as they already exist in parent class ClassInfo.


Himaswin:
Hi Mahmoud,

Thanks for the suggestion. I’ll try to add tests for isSealed and
isPermittedSubclass() to verify that parsing the .class file correctly
populates those fields. I’ll reach out if I have any doubts, if that’s okay.

cyrille-artho:
Thank you for your PR. We have already removed these fields in branch `java-17`.

PR #608, opened 2026-03-06
Fix: fields shadowing with sealed classes.

Mo-Naga:
In issue #605, the `JVMClassInfo ` has a shadowing fields: `permittedSubclassNames`, `isSealed`, that are exist in parent class `ClassInfo`. Removing this redefined field solves the problem.


cyrille-artho:
Thank you for your PR. We have already removed these two lines in branch `java-17`.

PR #614, opened 2026-03-31
fix: normalize permitted subclass names to dot notation in setPermittedSubclass

Darshan-dev57:
Fixes part of #605

setPermittedSubclass() was storing class names in slash notation
(e.g. gov/nasa/jpf/...) instead of dot notation (gov.nasa.jpf...).

This bug was pointed out in the issue comments by @Mahmoud-Khawaja.

The fix applies Types.getClassNameFromTypeName() which is the same
approach already used in setInterface() for consistency.


Mahmoud-Khawaja:
I will close it since you're working on #620

PR #617, opened 2026-04-23
Add missing java.lang.reflect.RecordComponent model class to java-17 branch

HritikRaj2:
### Description
While exploring the experimental `java-17` branch, I noticed that the `java.lang.reflect.RecordComponent` model class was missing from the `java.base` module. 

Since Java 14+, this class is required by the standard library to represent record components. This PR introduces the foundational JPF model class for `RecordComponent` with the necessary internal fields (`clazz`, `name`, `type`, `accessor`, etc.) that align with the standard OpenJDK specification. 

This acts as a necessary stepping stone for future JVM native peer implementations to correctly reflect over Record types in JPF without throwing `ClassNotFound` or missing definition errors.


cyrille-artho:
Thank you for your contribution.
As of now, the built-in Java library class `RecordComponent` works well for us, so we do not need an additional model class.

PR #620, opened 2026-06-12
Add parsing test for sealed classes and PermittedSubclasses attribute (#605)

Darshan-dev57:
## Summary

Adds the sealed class parsing test requested in #605, as discussed with @Mahmoud-Khawaja.

The test uses the `NonResolvedClassInfo` approach from the existing `ClassInfoTest` — it loads compiled `.class` files directly and verifies the parsed metadata, without needing a full JPF run (the runtime approach discussed in #608 does not work for this).

## Test cases (5)

- A sealed interface is detected (`isSealed()` returns `true`)
- A sealed abstract class is detected
- `PermittedSubclasses` contains the expected permitted subclass names (in dot notation)
- A non-permitted class is rejected by `isPermittedSubclass()`
- A plain (non-sealed) class is not detected as sealed

## Current results

On the current `java-17` branch, **3 of 5 tests fail**, confirming the field shadowing bug described in #605. I verified locally that all 5 tests pass once the duplicate `isSealed` and `permittedSubclassNames` fields are removed from `JVMClassInfo` (the fix proposed in #607/#608).

Note: this branch includes the dot-notation normalization from #614, which the permitted-name assertions depend on.

## Related

- Issue: #605
- Fix PRs: #607, #608
- Dot notation fix: #614

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
@@ -65,7 +65,7 @@ public class JVMClassInfo extends ClassInfo {
 
     @Override
     public void setPermittedSubclass(ClassFile cf, Object tag, int index, String subclassName) {
-      permittedSubclassNames[index] = subclassName;
+      permittedSubclassNames[index] = Types.getClassNameFromTypeName(subclassName);
     }
 
     @Override
--- a/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
@@ -43,8 +43,7 @@ public class JVMClassInfo extends ClassInfo {
   // To store partially resolved classes in setBootstrapMethod
   protected static HashMap resolvedClasses = new HashMap<String, JVMClassInfo>();
   protected ClassFile classFile;
-  protected String[] permittedSubclassNames;
-  protected boolean isSealed;
+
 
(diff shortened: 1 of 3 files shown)


Mahmoud-Khawaja:
Hi Darshan,
Thank you for the pr. Good work. tests are solid now. Can you also please fix the field shadowing?

Darshan-dev57:
Done — added the field shadowing fix in the latest commit. All 5 parsing tests pass.

src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
```diff
@@ -43,8 +43,7 @@ public class JVMClassInfo extends ClassInfo {
   // To store partially resolved classes in setBootstrapMethod
   protected static HashMap resolvedClasses = new HashMap<String, JVMClassInfo>();
   protected ClassFile classFile;
-  protected String[] permittedSubclassNames;
-  protected boolean isSealed;
+
```
cyrille-artho, line 46: Please avoid introducing new empty lines.

cyrille-artho (changes requested):
Great job! Just a minor cosmetic detail to change, then we can merge the PR. Please update it.

src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
Darshan-dev57: Done — removed the extra blank line.

cyrille-artho (approved):

PR #622, opened 2026-06-26
Fix: Missing check for STATIC method when executing (#250)

### LINKED ISSUE #250: Missing check for STATIC method when executing

BitsToBytes-Saksham:
Fixes #250.

This PR adds a check for `!callee.isStatic()` in `INVOKESTATIC.java` and throws an `IncompatibleClassChangeError`. Also includes a test `InvokeStaticTest`.

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/bytecode/INVOKESTATIC.java
+++ b/src/main/gov/nasa/jpf/jvm/bytecode/INVOKESTATIC.java
@@ -89,6 +89,11 @@ public class INVOKESTATIC extends JVMInvokeInstruction {
       return ti.createAndThrowException("java.lang.NoSuchMethodException", cname + '.' + mname);
     }
 
+    if (!callee.isStatic()) {
+      return ti.createAndThrowException("java.lang.IncompatibleClassChangeError",
+                                        "Expected static method " + callee.getFullName());
+    }
+
     // this can be actually different than (can be a base)
     ClassInfo ciCallee = callee.getClassInfo();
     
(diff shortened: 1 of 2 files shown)


BitsToBytes-Saksham:
Hi @cyrille-artho,

I have submitted a PR to resolve issue #250.

Changes Included:
- Bug Fix: Added a !callee.isStatic() check within INVOKESTATIC.java to correctly intercept non-static methods and throw an IncompatibleClassChangeError.
- Testing: Implemented InvokeStaticTest.java with automated regression tests, fully conforming to the TestJPF execution guidelines.

Could you please approve the workflow run and review the PR at your convenience. I am available to address any feedback or required adjustments.

Thank you for your time!

cyrille-artho:
Thank you for your contribution. In the current version of JPF, 5 unit tests fail with this patch. So we can unfortunately not merge this PR.

PR #623, opened 2026-07-01
fix: implement Class.isRecord() to call native ClassInfo.isRecord()

Darshan-dev57:
Fixes Class.isRecord() which was hardcoded to return false, breaking reflective record detection.

Changes:
- Added native isRecord____Z() method in JPF_java_lang_Class.java
- Changed Class.isRecord() to native in Class.java to delegate to ClassInfo.isRecord()

This allows proper record detection during reflection in Java 17 support.


Darshan-dev57:
Superseded by #625

PR #625, opened 2026-07-02
Fix: Record component generic signatures and Class.isRecord()

### LINKED ISSUE #624: Record components lose their generic signature (and annotations) during class file parsing
I wanted to check that record parsing keeps everything it should, so I tried a record with generic components:

```java
record Box<T>(T value, List<String> items) {}
```

I parsed it with JPF and printed the name, descriptor and signature of each component. The names (value, items) and descriptors (Ljava/lang/Object; for the erased T, and Ljava/util/List;) were all fine, but both signatures came back null.

To make sure the class file wasn't the problem, I ran javap -v on it. The Signature attributes are right there:

- value: TT;
- items: Ljava/util/List<Ljava/lang/String;>;

So the information is in the file. JPF reads it while parsing and then drops it.

## What's going on

Record parsing lives in JVMClassInfo, in the inner Initializer class. Two callbacks matter here.

setRecordComponent(...) runs once per component and builds the RecordComponentInfo with null for signature and annotations:

```java
RecordComponentInfo rci = new RecordComponentInfo(name, descriptor, null, null, null);
```

setRecordComponentAttribute(...) runs for each attribute of a component (Signature, the annotation attributes, etc). For a Signature attribute it does roughly this:

```java
cf.parseSignatureAttr(this, rci);   // parse, and pass rci in to be filled
signature = rci.getSignature();     // read it back
...
// rebuild the component from whatever was read back
recordComponents[componentIndex] = new RecordComponentInfo(name, descriptor, signature, annotations, typeAnnotations);
```

The catch is in how parseSignatureAttr hands the value back. It calls the shared setSignature handler, and that handler only stores the value if the object it was given is the expected type:

```java
public void setSignature(ClassFile cf, Object tag, String signature) {
  if (tag instanceof GenericSignatureHolder) {
    ((GenericSignatureHolder) tag).setGenericSignature(signature);
  }
}
```

Annotations work the same way. They only get stored if the object is an InfoObject:

```java
if (tag instanceof InfoObject) {
  ((InfoObject) tag).addAnnotations(...);
}
```

RecordComponentInfo is a plain class. It is not a GenericSignatureHolder and it is not an InfoObject, so both instanceof checks are false. The value has nowhere to go and gets dropped, with no error. setRecordComponentAttribute then reads the signature back (still null) and rebuilds the component with null.

The reason this works for fields and methods is that FieldInfo and MethodInfo already extend InfoObject and implement GenericSignatureHolder. Record components were just never hooked into that.

## How to reproduce

Parse a record with a generic component and check the components that come out:

```java
ClassInfo ci = /* parse Box.class, e.g. with the NonResolvedClassInfo used in the unit tests */;
RecordComponentInfo[] comps = ci.getRecordComponents();

comps[0].getName();        // "value"
comps[0].getDescriptor();  // "Ljava/lang/Object;"
comps[0].getSignature();   // null, should be "TT;"

comps[1].getSignature();   // null, should be "Ljava/util/List<Ljava/lang/String;>;"
```

A regression test fits well next to ClassInfoTest, using NonResolvedClassInfo so it parses a real compiled .class file without needing a full JPF run.

The fix is to make record components follow the same processing path as fields and methods.

Make RecordComponentInfo extend InfoObject and implement GenericSignatureHolder, so it can store generic signatures and reuse the existing annotation storage provided by InfoObject.
In JVMClassInfo, create the RecordComponentInfo object once in setRecordComponent, then pass that same instance to setRecordComponentAttribute so the parser updates it in place instead of rebuilding it.

This ensures that instanceof checks in the signature and annotation handling code succeed, allowing generic signatures and annotations to be stored correctly instead of being discarded.

The generic signature issue was reproduced directly. Annotation loss appears to have the same root cause (instanceof InfoObject), but wasn't independently reproduced because doing so requires a class loader not available in the lightweight parsing test.

### 1. Class.isRecord() always returns false

The modeled java.lang.Class.isRecord() is hardcoded:

```java
public boolean isRecord() {
  return false;
}
```

It never checks the parsed class info, so reflection inside a verified program thinks records are not records. ClassInfo already knows the answer (ClassInfo.isRecord() is set correctly during parsing), the model just doesn't ask it.

we can make isRecord() native in the model and let the peer delegate to ClassInfo.isRecord(), the same way isInterface() does.

In src/classes/modules/java.base/java/lang/Class.java, replace the hardcoded body with:

```java
public native boolean isRecord();
```

In src/peers/gov/nasa/jpf/vm/JPF_java_lang_Class.java, add the peer method next to isInterface____Z:

```java
@MJI
public boolean isRecord____Z (MJIEnv env, int robj){
  ClassInfo ci = env.getReferredClassInfo(robj);
  return ci.isRecord();
}
```

I tried this on JDK 17. Before the change, a record running inside verifyNoPropertyViolation prints:

```
Point.class.isRecord() = false
```

After the change it prints what you'd expect, and a non-record still reports false:

```
Point.class.isRecord() = true
String.class.isRecord() = false
```

### 2. No Class.getRecordComponents() support

This is a missing feature, not a bug. Even after fixing record component parsing, there is currently no modeled reflection support for record components, so verified programs cannot call Class.getRecordComponents().

The implementation would require:

Adding a modeled java.lang.reflect.RecordComponent class with the essential reflection APIs.
Modeling Class.getRecordComponents() as a native method.
Implementing the corresponding peer method to create and return RecordComponent objects from the parsed RecordComponentInfo data, similar to how reflected fields are handled.

The most complex part is implementing getAccessor(), which must return the corresponding reflected Method object.

Darshan-dev57:
fix #624

Addresses two bugs and one missing feature noted by @Mahmoud-Khawaja:

**1: fix Class.isRecord()**
- Class.isRecord() was hardcoded to return false
- Now correctly delegates to ClassInfo.isRecord()

**2: preserve record component generic signatures**
- Generic signatures and annotations were silently dropped during classfile parsing
- Fixed by implementing setRecordComponentAttribute properly in JVMClassInfo.Initializer

**3: add Class.getRecordComponents() reflection support**
- New modeled java.lang.reflect.RecordComponent class
- Class.getRecordComponents() peer implementation
- JPF_java_lang_reflect_RecordComponent peer with getName(), getType(), getAccessor(), getAnnotations(), getDeclaredAnnotations()
- getAccessor() fully implemented via JPF_java_lang_reflect_Method.createMethodObject()
- Annotations wired through env.newAnnotationProxies() backed by RecordComponentInfo.getAnnotations()
- 7 tests covering count, names, types, accessor, annotations, and non-record null return

All tests pass, no regressions.

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
@@ -251,12 +251,10 @@ public class JVMClassInfo extends ClassInfo {
       if (attrName.equals(ClassFile.SIGNATURE_ATTR)) {
         cf.parseSignatureAttr(this, rci);
       }
-      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_ANNOTATIONS_ATTR)
-              || attrName.equals(ClassFile.RUNTIME_INVISIBLE_ANNOTATIONS_ATTR)) {
+      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_ANNOTATIONS_ATTR)) {
         cf.parseAnnotationsAttr(this, rci);
       }
-      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_TYPE_ANNOTATIONS_ATTR)
-              || attrName.equals(ClassFile.RUNTIME_INVISIBLE_TYPE_ANNOTATIONS_ATTR)) {
+      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_TYPE_ANNOTATIONS_ATTR)) {
         cf.parseTypeAnnotationsAttr(this, rci);
       }
     }
--- a/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+++ b/src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
@@ -246,39 +246,19 @@ public class JVMClassInfo extends ClassInfo {
     public void setRecordComponentAttribute(ClassFile cf, Object tag, int componentIndex, int attrIndex, String attrName, int attrLength) {
       RecordComponentInfo rci = JVMClassInfo.this.recordComponents[componentIndex];
 
-      String name = rci.getName();
-      String descriptor = rci.getDescriptor();
-      String signature = rci.getSignature();
-      AnnotationInfo[] annotations = rci.getAnnotations();
-      TypeAnnotationInfo[] typeAnnotations = rci.getTypeAnnotations();
-
+      // RecordComponentInfo now extends InfoObject and implements GenericSignatureHolder,
+      // so the parse callbacks populate rci directly instead of silently no-op'ing.
       if (attrName.equals(ClassFile.SIGNATURE_ATTR)) {
-        // Use the signature consumer pattern
         cf.parseSignatureAttr(this, rci);
-        signature = rci.getSignature();
       }
-      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_ANNOTATIONS_ATTR)) {
-        annotations = new AnnotationInfo[0]; // Initialize or preserve existing
+      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_ANNOTATIONS_ATTR)
+              || attrName.equals(ClassFile.RUNTIME_INVISIBLE_ANNOTATIONS_ATTR)) {
         cf.parseAnnotationsAttr(this, rci);
-        annotations = rci.getAnnotations();
       }
-      else if (attrName.equals(ClassFile.RUNTIME_INVISIBLE_ANNOTATIONS_ATTR)) {
-        cf.parseAnnotationsAttr(this, rci);
-        // we might need to merge with existing annotations
-      }
-      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_TYPE_ANNOTATIONS_ATTR)) {
-        // Set up for type annotation parsing
-        typeAnnotations = new TypeAnnotationInfo[0];
-        cf.parseTypeAnnotationsAttr(this, rci);
-        typeAnnotations = rci.getTypeAnnotations();
-      }
-      else if (attrName.equals(ClassFile.RUNTIME_INVISIBLE_TYPE_ANNOTATIONS_ATTR)) {
+      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_TYPE_ANNOTATIONS_ATTR)
+              || attrName.equals(ClassFile.RUNTIME_INVISIBLE_TYPE_ANNOTATIONS_ATTR)) {
         cf.parseTypeAnnotationsAttr(this, rci);
       }
-
-      // a new RecordComponentInfo with all updated attributes
-      JVMClassInfo.this.recordComponents[componentIndex] =
-              new RecordComponentInfo(name, descriptor, signature, annotations, typeAnnotations);
     }
 
     @Override
--- a/src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_RecordComponent.java
+++ b/src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_RecordComponent.java
@@ -18,6 +18,9 @@ import gov.nasa.jpf.vm.MJIEnv;
 import gov.nasa.jpf.vm.NativePeer;
 import gov.nasa.jpf.vm.RecordComponentInfo;
 import gov.nasa.jpf.vm.ThreadInfo;
+import gov.nasa.jpf.vm.AnnotationInfo;
+import gov.nasa.jpf.vm.MethodInfo;
+import gov.nasa.jpf.vm.JPF_java_lang_reflect_Method;
 import gov.nasa.jpf.vm.Types;
 
 /**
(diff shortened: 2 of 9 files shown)


Mahmoud-Khawaja:
Hey Darshan,
The two bug fixes and the reflection feature work. I ran the branch and verified signatures, isRecord(), getType(), and parse-time annotation storage. But three things before this can merge: 

1-retarget the PR to java-17 it currently sits on call-site-generation 

2- getAccessor() is declared native with no peer, so calling it crashes with UnsatisfiedLinkError. noting it as "follow-up" in the description doesn't help; remove the declaration or throw UnsupportedOperationException; 

3- the annotations half of [#624](https://github.com/javapathfinder/jpf-core/issues/624) is only fixed at parse time. getAnnotations() hardcodes an empty array, so annotations are still invisible through reflection, and there's no annotation test anywhere in the PR. Please wire annotations through the peer (see how JPF_java_lang_Class.getAnnotations uses env.newAnnotationProxies) and add a test with an annotated component.

Darshan-dev57:
Hi @Mahmoud-Khawaja — addressed all three review points. Rebased onto java-17 , `getAccessor()` fully implemented and tested, annotations wired through `env.newAnnotationProxies()` with an annotated component test. All 7 reflection tests pass.

src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
```diff
@@ -246,39 +246,19 @@ public void setRecordComponent(ClassFile cf, Object tag, int index, String name,
...
-        signature = rci.getSignature();
-      }
-      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_ANNOTATIONS_ATTR)) {
-        annotations = new AnnotationInfo[0]; // Initialize or preserve existing
-        cf.parseAnnotationsAttr(this, rci);
-        annotations = rci.getAnnotations();
       }
-      else if (attrName.equals(ClassFile.RUNTIME_INVISIBLE_ANNOTATIONS_ATTR)) {
+      else if (attrName.equals(ClassFile.RUNTIME_VISIBLE_ANNOTATIONS_ATTR)
+              || attrName.equals(ClassFile.RUNTIME_INVISIBLE_ANNOTATIONS_ATTR)) {
```
Mahmoud-Khawaja: this puts invisible annotations into the same InfoObject.annotations array as the visible ones, so a CLASS-retention annotation would now show up in getAnnotations() at runtime. setFieldAttribute only parses the VISIBLE attrs for exactly this reason. drop the invisible branches (same for the invisible type annotations one).

src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_RecordComponent.java
```diff
@@ -0,0 +1,125 @@
...
+    if (classRef == MJIEnv.NULL) return MJIEnv.NULL;
+    ClassInfo recordCI = env.getReferredClassInfo(classRef);
+
+    // accessor method has same name as component, no args, returns component type
+    String accessorSignature = rci.getName() + "()" + rci.getDescriptor();
+    MethodInfo mi = recordCI.getMethod(accessorSignature, false);
+    if (mi == null) return MJIEnv.NULL;
+
+    // get ClassInfo for java.lang.reflect.Method (needed by createMethodObject)
+    ClassInfo methodCI = ClassLoaderInfo.getSystemResolvedClassInfo("java.lang.reflect.Method");
```
Mahmoud-Khawaja: resolved != initialized. If nothing touched Method before this call, createMethodObject can fail. Use the same pattern as getEnclosingConstructor in JPF_java_lang_Class: getInitializedClassInfo(env, ...) + env.repeatInvocation() when it returns null.

src/peers/gov/nasa/jpf/vm/JPF_java_lang_reflect_RecordComponent.java
```diff
@@ -0,0 +1,125 @@
...
+    return sig != null ? env.newString(sig) : MJIEnv.NULL;
+  }
+
+  @MJI
+  public int getAnnotations_____3Ljava_lang_annotation_Annotation_2(MJIEnv env, int objRef) {
+    RecordComponentInfo rci = getRecordComponentInfo(env, objRef);
+    if (rci == null) return env.newObjectArray("Ljava/lang/annotation/Annotation;", 0);
+    AnnotationInfo[] ai = rci.getAnnotations();
+    if (ai == null || ai.length == 0) return env.newObjectArray("Ljava/lang/annotation/Annotation;", 0);
+    return env.newAnnotationProxies(ai);
```
Mahmoud-Khawaja: newAnnotationProxies throws ClinitRequired if the proxy class isn't initialized yet. Every other caller wraps it, see getAnnotations in JPF_java_lang_Class:
``` java
try {
  return env.newAnnotationProxies(ai);
} catch (ClinitRequired x) {
  env.handleClinitRequest(x.getRequiredClassInfo());
  return MJIEnv.NULL;
}
```
same for getDeclaredAnnotations below. Your test passes because the annotation class happens to be initialized by that point, but this will crash on first use in a real run

src/tests/java17/records/RecordComponentReflectionTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Test
+  public void testGetRecordComponents_annotations() {
+    if (verifyNoPropertyViolation()) {
+      RecordComponent[] components = AnnotatedPoint.class.getRecordComponents();
+      assertNotNull(components);
+      assertEquals(2, components.length);
+      java.lang.annotation.Annotation[] annotations = components[0].getAnnotations();
+      assertNotNull(annotations);
+      assertTrue(annotations.length > 0);
```
Mahmoud-Khawaja, line 89: this only proves the array isn't empty. Assert the actual content:
``` java
assertEquals(ComponentLabel.class, annotations[0].annotationType());
assertEquals("x-coord", ((ComponentLabel) annotations[0]).value());
```
that's what actually exercises the proxy. and it's where the ClinitRequired issue from my other comment would show up.

Mahmoud-Khawaja:
Hi Darshan,
please check my comments. feel free to reach out if you have any questions.

cyrille-artho (changes requested):
Thank you very much for your PR!
Please check Mahmoud's comments and change the code to address them.

Darshan-dev57:
Hi @Mahmoud-Khawaja — addressed all four review points in commit 520f16d.

cyrille-artho (changes requested):
Looks good to me, but let's wait until Mahmoud has checked your changes.
I would change the first two comment lines, updating the (C) year to 2026 and using your name (instead of NASA) as the copyright holder in cases where you have created a new file.

Mahmoud-Khawaja:
Looks good to me. I think we can merge this.

cyrille-artho (approved):

PR #627, opened 2026-07-25
Fix boxed Float and Double string concatenation (#626)

### LINKED ISSUE #626: Boxed Float and Double break string concatenation (BootstrapMethodError: not a int/long field: value)
I was going over the string concat work on the call-site-generation branch and wanted to make sure every operand type that can reach makeConcatWithConstants actually works, not just the ones we already have tests for. The primitive types and String were fine, so I started trying the boxed wrapper types one by one.

Integer and Long were fine. But the moment I concatenated a boxed Float, the whole run blew up:

```java
public class One {
    Float f = 1.5f;
    String s = "x=" + f;
    System.out.println("result = " + s);
  }
}
```

Running this under JPF:

```
One.main()

====================================================== search started: 7/21/26, 4:24 AM

====================================================== error 1
java.lang.BootstrapMethodError: CallSite execution failed: not a int field: value
	at One.main(One.java:4)

====================================================== snapshot #1
thread java.lang.Thread:{id:0,name:main,status:RUNNING,priority:5,isDaemon:false,lockCount:0,suspendCount:0}
  call stack:
	at One.main(One.java:4)

error #1: gov.nasa.jpf.vm.NoUncaughtExceptionsProperty "java.lang.BootstrapMethodError: CallSite execution..."
```

Doing the same thing with a boxed Double fails the same way, just with a different field type in the message:

```
====================================================== error 1
java.lang.BootstrapMethodError: CallSite execution failed: not a long field: value
```

So it is only boxed Float and Double. Integer, Long, Byte, Short, Character and Boolean (all boxed) work, and the primitive float and double work too. This is plain everyday Java ("x=" + someFloatObject), so JPF is diverging from the JVM on ordinary programs.

## Why boxed types even reach this code

At first I expected javac to convert the object to a String before the invokedynamic, the way it does for arbitrary objects. But that is not what happens for the wrapper types. If you look at the bytecode, a boxed Float is passed straight into makeConcatWithConstants:

```
static java.lang.String boxed(java.lang.Float);
  Code:
     0: aload_0
     1: invokedynamic #0:makeConcatWithConstants:(Ljava/lang/Float;)Ljava/lang/String;
     6: areturn
```

So the Float object really does flow through our concat argument handling as an ElementInfo, and we read its backing value field. That read is where it breaks.

## What's going on

The argument handling reads the wrapper's value field with the wrong declared type, then does an extra bit reinterpretation on top.

In src/main/gov/nasa/jpf/jvm/bytecode/INVOKEDYNAMIC.java, convertBoxedPrimitive() (this is the one that actually gets hit, while popping the arguments):

```java
case "java.lang.Float":   return Float.intBitsToFloat(ei.getIntField("value"));
case "java.lang.Double":  return Double.longBitsToDouble(ei.getLongField("value"));
```

The same thing is duplicated in src/main/gov/nasa/jpf/vm/JPFStringConcatHelper.java, handleElementInfo():

```java
case "java.lang.Float":
    sb.append(Float.intBitsToFloat(ei.getIntField("value"))); break;
case "java.lang.Double":
    sb.append(Double.longBitsToDouble(ei.getLongField("value"))); break;
```

But java.lang.Float.value is declared float, and java.lang.Double.value is declared double. ElementInfo.getIntField / getLongField check the declared field type and throw "not a int field: value" / "not a long field: value". That is exactly the message in the error above.

I think the intBitsToFloat / longBitsToDouble came from copying the primitive path in popPrimitiveArgument, where it is correct:

```java
} else if (type == float.class) {
  return Float.intBitsToFloat(frame.pop());
```

There the raw operand stack slot genuinely holds the int/long bit pattern, so the reinterpretation makes sense. But a heap wrapper object already stores a real float/double in its value field, so there is nothing to reinterpret and reading it as int/long is just wrong. Integer and Long happen to work only because their value fields really are int and long.

## How to reproduce

```
git checkout call-site-generation
JAVA_HOME=/path/to/jdk-17 ./gradlew buildJars
```

Compile the One.java above, then:

```
printf 'target=One\nclasspath=/tmp/diff\n' > /tmp/diff/one.jpf
java -jar build/RunJPF.jar /tmp/diff/one.jpf
```

Expected: result = x=1.5, no errors.
Actual: BootstrapMethodError: CallSite execution failed: not a int field: value.

Swap Float for Double to get the "not a long field: value" variant.

what i recommed to fix Read each wrapper's field with its real declared type and drop the bit conversion, in both places.

INVOKEDYNAMIC.convertBoxedPrimitive():

```java
case "java.lang.Float":   return ei.getFloatField("value");
case "java.lang.Double":  return ei.getDoubleField("value");
```

JPFStringConcatHelper.handleElementInfo():

```java
case "java.lang.Float":   sb.append(ei.getFloatField("value")); break;
case "java.lang.Double":  sb.append(ei.getDoubleField("value")); break;
```

ElementInfo already has getFloatField and getDoubleField. I tried this locally and after the change all the wrapper types concatenate correctly. I checked finite values, NaN and Infinity, and null wrappers, and the output matches the real JVM (i=42, l=9999999999, d=3.14, f=1.5, then NaN/-Infinity, then null,null,null).

A regression test in src/tests/java11 next to the existing string concat tests would fit well, something like:

```java
@Test
public void testBoxedFloatAndDoubleConcat() {
    if (verifyNoPropertyViolation()) {
        Float f = 1.5f;
        Double d = 3.14;
        assertEquals("f=1.5,d=3.14", "f=" + f + ",d=" + d);
        assertEquals("NaN/-Infinity", "" + Double.NaN + "/" + Float.NEGATIVE_INFINITY);
    }
}
```

Darshan-dev57:
## Summary

Fixes #626 boxed `Float` and `Double` string concatenation by reading wrapper fields using their correct types (`getFloatField()`/`getDoubleField()`) instead of `getIntField()`/`getLongField()`.

## Changes

* Updated `INVOKEDYNAMIC.convertBoxedPrimitive()`
* Updated `JPFStringConcatHelper.handleElementInfo()`
* Added `StringConcatBoxedFloatDoubleTest` as a regression test.

## Testing

* Verified the new test fails before the fix.
* Verified all 7 tests pass after the fix.

### DIFF
--- a/src/main/gov/nasa/jpf/jvm/bytecode/INVOKEDYNAMIC.java
+++ b/src/main/gov/nasa/jpf/jvm/bytecode/INVOKEDYNAMIC.java
@@ -297,8 +297,8 @@ public class INVOKEDYNAMIC extends Instruction {
       case "java.lang.Short":   return ei.getShortField("value");
       case "java.lang.Integer": return ei.getIntField("value");
       case "java.lang.Long":    return ei.getLongField("value");
-      case "java.lang.Float":   return Float.intBitsToFloat(ei.getIntField("value"));
-      case "java.lang.Double":  return Double.longBitsToDouble(ei.getLongField("value"));
+      case "java.lang.Float":   return ei.getFloatField("value");
+      case "java.lang.Double":  return ei.getDoubleField("value");
       case "java.lang.Boolean": return ei.getBooleanField("value");
       case "java.lang.String":  return ei.asString();
       default: return ei;
--- a/src/main/gov/nasa/jpf/vm/JPFStringConcatHelper.java
+++ b/src/main/gov/nasa/jpf/vm/JPFStringConcatHelper.java
@@ -122,9 +122,9 @@ public class JPFStringConcatHelper {
             case "java.lang.Integer": sb.append(ei.getIntField("value")); break;
             case "java.lang.Boolean": sb.append(ei.getBooleanField("value")); break;
             case "java.lang.Float":
-                sb.append(Float.intBitsToFloat(ei.getIntField("value"))); break;
+                sb.append(ei.getFloatField("value")); break;
             case "java.lang.Double":
-                sb.append(Double.longBitsToDouble(ei.getLongField("value"))); break;
+                sb.append(ei.getDoubleField("value")); break;
             case "java.lang.Long":    sb.append(ei.getLongField("value")); break;
             default: sb.append(ei.toString()); break;
         }
--- a/src/tests/java11/StringConcatBoxedFloatDoubleTest.java
+++ b/src/tests/java11/StringConcatBoxedFloatDoubleTest.java
@@ -1,0 +1,67 @@
+package java11;
+
+import gov.nasa.jpf.util.test.TestJPF;
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+
+public class StringConcatBoxedFloatDoubleTest extends TestJPF {
+
+  @Test
+  public void testBoxedFloatConcat() {
+    if (verifyNoPropertyViolation()) {
+      Float f = 1.5f;
+      String result = "f=" + f;
+      assertEquals("f=1.5", result);
+    }
+  }
+
+  @Test
+  public void testBoxedDoubleConcat() {
+    if (verifyNoPropertyViolation()) {
+      Double d = 3.14;
+      String result = "d=" + d;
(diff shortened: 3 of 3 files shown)


cyrille-artho (approved):

cyrille-artho:
Looks good to me! Mahmoud, any suggestions/comments?

Mahmoud-Khawaja:
All good! I think we can merge that.

PR #630, opened 2026-08-11
# Fix #629: Incorrect method reference results when return value requires boxing

### LINKED ISSUE #629: Method references return wrong values when the result needs boxing
I hit this while I was testing records for #628, but it has nothing to do with that branch. it reproduces on current master too.

Basically, a method reference gives back garbage whenever the method returns a primitive but the functional interface wants an object. So `Function<String,Integer> f = String::length` is broken, while `ToIntFunction<String> f = String::length` is fine. It took me a while to narrow it down because nothing throws — you just silently get a wrong value back.

```java
import java.util.function.*;
public class MethodRefBoxingRepro {
    ToIntFunction<String> f1 = String::length;    // primitive return, no boxing
    System.out.println("ToIntFunction String::length -> " + f1.applyAsInt("hello"));
    Function<String,String> f2 = String::trim;    // reference return, no boxing
    System.out.println("Function String::trim -> [" + f2.apply(" hi ") + "]");
    Function<String,Integer> f3 = String::length; // needs boxing
    System.out.println("Function String::length -> " + f3.apply("hello"));
  }
}
```

On a real JVM:

```
$ java MethodRefBoxingRepro
ToIntFunction String::length -> 5
Function String::trim -> [hi]
Function String::length -> 5
```

Under JPF:

```
$ java -jar build/RunJPF.jar +classpath=. MethodRefBoxingRepro
ToIntFunction String::length -> 5
Function String::trim -> [hi]
Function String::length -> interface java.lang.reflect.AnnotatedElement
```

Yes, `"hello".length()` is an interface now 

My guess from the outside: the primitive return value gets treated as a heap reference. `length()` returns 5, and what comes back looks like whatever object happens to live at address 5. The other variants I tried fit that pattern too:

```
$ java -jar build/RunJPF.jar +classpath=. MethodRefRepro
String::length -> interface java.lang.reflect.AnnotatedElement
s::length      -> java.lang.reflect.AnnotatedElement
Integer::parseInt -> java.lang.Thread
Box::get       -> java.lang.reflect.GenericDeclaration
String::isEmpty -> java.lang.ClassLoader@43980
```

Expected 5 / 6 / 123 / 9 / true. Note `Integer.parseInt("123")` coming back as a `Thread` — a bigger number lands further into the heap.

So it's not one specific kind of method reference: unbound (`String::length`), bound (`s::length`), static (`Integer::parseInt`) and user classes (`Box::get`) all do this. Regular lambdas like `x -> x.length()` work, I assume because javac puts the lambda body in a synthetic method that already returns `Integer`.

The annoying part is how common this shape is — `Function`, `Supplier`, `Predicate` are all generic, so something as ordinary as `stream.map(String::length)` gets silently wrong data, and JPF keeps exploring with it instead of failing.

alwaysalearner1234:
## Description

Fixes #629.

Method references that return a primitive value were returning incorrect values when the functional interface expected an object.

For example:

```java
Function<String, Integer> f = String::length;
```

The primitive `int` result was not being boxed before being returned as an object.

### Fix

* Detect primitive-to-reference return type conversions.
* Box primitive values using the appropriate wrapper `valueOf()` method.
* Keep existing primitive and reference return behavior unchanged.

### Example

Before:

```text
Function<String, Integer> f = String::length;
→ incorrect value
```

After:

```text
Function<String, Integer> f = String::length;
→ 5
```

### Testing

Tested with method references including:

* `String::length`
* `String::isEmpty`
* `Integer::parseInt`
* Bound instance methods returning primitives
* `ToIntFunction<String>` to ensure existing behavior still works

### DIFF
--- a/bug_investigation_report.md
+++ b/bug_investigation_report.md
@@ -1,0 +1,68 @@
+# Issue #629 Investigation and PR #630 Report
+
+## Root Cause
+When a method reference targets a method returning a primitive type, but the functional interface expects a reference type (e.g., `Function<String, Integer> f = String::length`), Java's `LambdaMetafactory` generates an adapter class that automatically boxes the primitive return value into a wrapper object before returning it. 
+Prior to PR #630, JPF short-circuited `LambdaMetafactory` by generating a synthetic `DirectCall` method directly from the `implMethod` (e.g. `String.length()I`). It incorrectly inspected only the functional interface's expected return type, ignoring the actual primitive return type of the underlying method. It generated an `areturn` instruction, assuming an object reference was on the operand stack. However, since the underlying method returned a primitive, a raw primitive value was on the stack. JPF treated this primitive value (e.g., `5`) as an object reference ID, leading to garbage object references or JVM crashes.
+
+## Implementation
+JPF processes method references by intercepting the `LambdaMetafactory` bootstrap method in `JVMClassInfo.java` (specifically inside `setLambdaDirectCallCode`). It synthesizes a new class and method that directly invoke the target method (`implMethod`) and returns the result. 
+To fix the boxing issue, the implementation now correctly compares the primitive return type of the callee method (`calleeReturnType`) with the reference return type of the expected SAM (`samReturnType`). If the callee returns a primitive (length 1, not 'V') and the SAM expects a reference (length > 1), JPF inserts an `invokestatic` call to the corresponding wrapper class's `valueOf` method (e.g., `java.lang.Integer.valueOf`) before executing the `areturn` instruction.
+
+## Files Changed
+File: src/main/gov/nasa/jpf/jvm/JVMClassInfo.java
+Method: setLambdaDirectCallCode(MethodInfo miDirectCall, BootstrapMethodInfo bootstrapMethod)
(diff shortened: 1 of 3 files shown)


cyrille-artho:
On my Mac laptop with OpenJDK 11.0.32, 20 unit tests fail.
The result is the same for the current version of jpf-core, branch `master`, in the main repository.
On Darshan's system (OpenJDK 11.0.31 on IntelliJ on Windows), all unit tests pass.

alwaysalearner1234:
Thanks for taking the time to test this and verify the behavior on your end!

I investigated the test failures locally and compared the PR branch against a clean checkout of `upstream/master`. The failures are also reproducible on the current `master` branch, so they do not appear to be introduced by the changes in this PR.

The failures seem to be environment-specific, particularly given the difference between the macOS/OpenJDK 11.0.32 and Windows/OpenJDK 11.0.31 environments.

For this PR, I focused on the method-reference boxing issue described in #629 and verified that the relevant regression cases behave correctly after the change.

I'll keep the PR focused on #629 and leave the pre-existing test failures as a separate issue.

cyrille-artho (changes requested):
Two unit tests fail.

PR #631, opened 2026-08-11
fix: correct record equals/toString/hashCode for reference and array components (#628)

### LINKED ISSUE #628: Record equals/toString don't match real JVM behavior for reference and array components
I was testing some edge cases on the `call-site-generation` branch and ran into three problems with records, all in cases where a record component is a reference type or an array. From the stack traces it looks related to the direct execution of record methods added in #554, but I haven't dug into the cause yet.

I compiled the repros with `javac --release 17` and ran them with `build/RunJPF.jar`.

## 1. `equals()` crashes on records with array components

Any record with an array component throws a `BootstrapMethodError` as soon as `equals()` has to compare two different arrays. On a real JVM this just returns `false`:

```java
public class RecordArrayRepro {
  record Wrap(int[] data) {}
    int[] shared = {1, 2, 3};
    Wrap w1 = new Wrap(shared);
    Wrap w2 = new Wrap(shared);
    Wrap w3 = new Wrap(new int[]{1, 2, 3});
    System.out.println("sameArray:  " + w1.equals(w2));
    System.out.println("equalArray: " + w1.equals(w3));
  }
}
```

```
$ java RecordArrayRepro
sameArray:  true
equalArray: false

$ java -jar build/RunJPF.jar +classpath=. RecordArrayRepro
sameArray:  true

====================================================== error 1
java.lang.BootstrapMethodError: CallSite execution failed: class gov.nasa.jpf.vm.DynamicElementInfo cannot be cast to class java.lang.Integer (gov.nasa.jpf.vm.DynamicElementInfo is in unnamed module of loader gov.nasa.jpf.JPFClassLoader @6f94fa3e; java.lang.Integer is in module java.base of loader 'bootstrap')
	at RecordArrayRepro$Wrap.equals(RecordArrayRepro.java:2)
	at RecordArrayRepro.main(RecordArrayRepro.java:9)
```

Interesting that the `sameArray` case works and only the comparison of two distinct arrays crashes.

## 2. `toString()` prints JPF's internal representation for reference components

String and boxed components print JPF heap info instead of the actual value:

```java
public class RecordStringRepro {
  record Person(String name, int age) {}
    Person p = new Person("alice", 30);
    System.out.println("toString: " + p);
  }
}
```

```
$ java RecordStringRepro
toString: Person[name=alice, age=30]

$ java -jar build/RunJPF.jar +classpath=. RecordStringRepro
toString: Person[name=java.lang.String@209, age=30]
```

Same story with boxed components — `record Holder(Integer val, int[] arr)` with `val=42`:

```
$ java RecordBoxedToStringRepro
toString: Holder[val=42, arr=[I@61bbe9ba]

$ java -jar build/RunJPF.jar +classpath=. RecordBoxedToStringRepro
toString: Holder[val=java.lang.Integer@2b7, arr=[I@30d]
```

Oddly enough, nested records print fine (`Outer[in=Inner[x=7], ...]`) — it's only the non-record reference components that come out wrong. Primitive components (like `age` above) are also fine.

## 3. `equals()` compares non-String reference components by identity

Records with boxed components compare by reference instead of by value:

```java
public class RecordBoxedRepro {
  record Holder(Integer val) {}
    Holder h1 = new Holder(Integer.valueOf(1000)); // outside the Integer cache
    Holder h2 = new Holder(Integer.valueOf(1000));
    System.out.println("equals: " + h1.equals(h2));
  }
}
```

```
$ java RecordBoxedRepro
equals: true

$ java -jar build/RunJPF.jar +classpath=. RecordBoxedRepro
equals: false
```

String components do compare by value, so it seems specific to other reference types. I'd expect this to hit anything with a value-based equals: boxed primitives, BigDecimal, user classes that override equals, etc.

## What still works

I also checked null components, nested record toString, and primitive concat args (char/boolean/long/byte/short/float) — those all match HotSpot. The existing test suite passes too, these cases just aren't covered by it.

@Darshan-dev57 please look into this issue and @cyrille-artho can help us with that

Darshan-dev57:
Fix #628 
Three bugs fixed in executeRecord() on call-site-generation branch:

1. equals() for array components: was doing deep element comparison, now uses reference identity (JEP 395: Objects.equals() on arrays delegates to Object.equals() = reference identity).

2. toString() for String/boxed/reference components: was printing JPF heap addresses (e.g. java.lang.String@209). Now:
   - String: ei.asString() for actual content
   - Boxed JDK types: wrapped value via 'value' field
   - Arrays: [I@hexHash format (Object.toString() semantics)
   - Other references: ClassName@hexHash

3. equals() for boxed types (Integer, Long, etc.): was using reference identity, now compares by wrapped value using a safe whitelist of known JDK boxed types. User classes with a 'value' field but identity-based equals() are NOT affected.

4. hashCode() made consistent with Java record semantics:

    - String: content hash
    - Boxed JDK types: wrapped value hash
    - Arrays: content-based hash, recursively handling multidimensional arrays
    - Nested records: recursive hash
    - Other references: identity-based hash

Added RecordReferenceComponentTest regression coverage for arrays, strings, boxed types, identity classes, null components, nested records, different-length arrays, and multidimensional arrays.

Verified: all 17 regression tests pass and the full build succeeds.

### DIFF
--- a/src/tests/java17/records/RecordReferenceComponentTest.java
+++ b/src/tests/java17/records/RecordReferenceComponentTest.java
@@ -15,6 +15,7 @@ public class RecordReferenceComponentTest extends TestJPF {
   record BoxedRecord(Integer val) {}
   record NullableRecord(String s, Integer i) {}
   record NestedRecord(StringRecord inner, int x) {}
+  record LongArrayRecord(long[][] data) {}
 
   static class IdentityClass {
     int value;
@@ -155,6 +156,38 @@ public class RecordReferenceComponentTest extends TestJPF {
       ArrayRecord r1 = new ArrayRecord(arr);
       ArrayRecord r2 = new ArrayRecord(arr);
       assertEquals(r1.hashCode(), r2.hashCode());
+    }
+  }
+
+  @Test
+  public void testHashCodeDifferentLengthArrays() {
+    if (verifyNoPropertyViolation()) {
+      int[] arr2a = {1, 2};
+      int[] arr2b = {1, 2};
+      int[] arr3 = {1, 2, 3};
+
+      ArrayRecord r2a = new ArrayRecord(arr2a);
+      ArrayRecord r2b = new ArrayRecord(arr2b);
+      ArrayRecord r3 = new ArrayRecord(arr3);
+
+      assertEquals(r2a.hashCode(), r2b.hashCode());
+      assertTrue(r2a.hashCode() != r3.hashCode());
+    }
+  }
+
+  @Test
+  public void testHashCodeHigherDimensionalArray() {
+    if (verifyNoPropertyViolation()) {
+      long[][] arr1 = {{1L, 2L}, {3L, 4L}};
+      long[][] arr2 = {{1L, 2L}, {3L, 4L}};
+
+      LongArrayRecord r1 = new LongArrayRecord(arr1);
+      LongArrayRecord r2 = new LongArrayRecord(arr2);
+
+      assertEquals(r1.hashCode(), r2.hashCode());
+
+      LongArrayRecord r3 = new LongArrayRecord(arr1);
+      assertEquals(r1.hashCode(), r3.hashCode());
     }
   }
 }
--- a/src/tests/java17/records/RecordReferenceComponentTest.java
+++ b/src/tests/java17/records/RecordReferenceComponentTest.java
@@ -1,0 +1,160 @@
+package java17.records;
+
+import gov.nasa.jpf.util.test.TestJPF;
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+/**
+ * Regression tests for GitHub issue #628:
+ * Record equals/toString/hashCode with reference and array components.
+ */
+public class RecordReferenceComponentTest extends TestJPF {
+
+  record ArrayRecord(int[] data) {}
+  record StringRecord(String name, int age) {}
+  record BoxedRecord(Integer val) {}
+  record NullableRecord(String s, Integer i) {}
+  record NestedRecord(StringRecord inner, int x) {}
+
+  static class IdentityClass {
+    int value;
+    IdentityClass(int v) { this.value = v; }
+  }
+  record IdentityRecord(IdentityClass obj) {}
+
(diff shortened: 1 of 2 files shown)


src/tests/java17/records/RecordReferenceComponentTest.java
```diff
@@ -0,0 +1,160 @@
...
+      assertEquals(r1.hashCode(), r2.hashCode());
+    }
+  }
+
+  @Test
+  public void testHashCodeConsistencyArray() {
+    if (verifyNoPropertyViolation()) {
+      int[] arr = {1, 2, 3};
+      ArrayRecord r1 = new ArrayRecord(arr);
+      ArrayRecord r2 = new ArrayRecord(arr);
```
cyrille-artho, line 157: Add a variant where you compare two arrays of length != 3, with the same content. Their hash code should be equal but different to the hash of the three-element arrays.
Add a variant where you compare two higher-dimensional arrays, e.g., long[][], where the hash code should again differ from the one-dimensional arrays but match another higher-dimensional array with the same contents.

cyrille-artho (approved):
Good work! Tests: Please check whether higher dimensions in arrays (as well as their content) are properly hashed.

cyrille-artho:
With the new tests, there are two possible outcomes: (1) they do not all pass, in which case please look at the hash values and how to fix them; (2) they pass, in which case please update the PR.

cyrille-artho (changes requested):
Please update the tests as in the previous comment.

Darshan-dev57:
Updated as requested. Added tests for different-length arrays and multidimensional arrays, including equivalent multidimensional arrays. The array hash implementation now recursively hashes array contents. All 17 RecordReferenceComponentTest tests pass, and the full build succeeds.

cyrille-artho:
Hi,
Thank you for the update. I have now 12 unit tests failing on Mac OS with Java 17.0.20.
With Java 17.0.9 on Mac OS (graalvm), I have 14 unit tests failing.
On Ubuntu 24.04.02 (Darshan's system), all unit tests pass.

cyrille-artho (approved):

